PR #18, opened 2023-06-12
PIC-2424 add unique constraint to CRN

cairnsp:
- Add unique constraint to CRN
- Update controller advice to return HTTP conflict
- Spring boot version upgrade and associated refactoring
- Integration test
- Fixed flyway issues (NB. will require a schema drop)


src/main/resources/db/migration/V1__add_person_table.sql
```diff
@@ -6,9 +6,45 @@ CREATE TABLE IF NOT EXISTS person
...
+CREATE TABLE revinfo
+(
+    rev INTEGER NOT NULL,
+    revtstmp bigint,
+    PRIMARY KEY (rev)
+);
+
+DROP TABLE IF EXISTS person_aud;
+
+CREATE TABLE person_aud
```
rschallapalli, line 35: CCS auto creates/updates audit tables. Isn't it a good idea?

rschallapalli (approved):

PR #38, opened 2023-11-22
CPR-70-data model

vb-moj:
. Added data model classes required to process hmcts SQS messages.


vb-moj:
> Looks good, and only have a couple of suggestions:
> 
>     * create an hmcts package to give :
>       ...personrecord.model.hmcts.commonplatform and:
>       ...personrecord.model.hmcts.libra
> 
>     * remove the 'CP' prefix from the common platform classes as the package name already specifies this
> 
>     * As a result of above, add the appropriate @JSON... annotation to specifiy the Pojo name

I also thought the same about the 'CP' prefix and am happy to discuss the suggestions.

PR #56, opened 2023-12-19
CPR-43-consume-delius-messages

vb-moj:
. Added missing else block.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/CourtCaseEventsListener.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/CourtCaseEventsListener.kt
@@ -54,7 +54,8 @@ class CourtCaseEventsListener(
           )
         }
       }
+    }else {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/OffenderDomainEventListener.kt
```diff
@@ -0,0 +1,53 @@
+package uk.gov.justice.digital.hmpps.personrecord.message.listeners
```
markreesmoj, line 1: Very minor but I note that the above class is called CourtCaseEvent**s**Listener  and this one is OffenderDomainEventListener - I feel we should go for either "Event" or "Events" and stick with it

vb-moj: 'Events' is the correct one here as well. Will update it in the next PR.Thanks.

PR #63, opened 2024-01-10
CPR-126

vb-moj:
. Added indexes on defendant and offender table.


src/main/resources/db/migration/V2__add_indexes.sql
```diff
@@ -0,0 +1,4 @@
+BEGIN;
+create index idx_defendant_crn_pnc on defendant (crn, pnc_number);
+create index idx_offender_crn_pnc on offender (crn, pnc_number);
+COMMIT;
```
vb-moj, line 4: Do we need to add name fields as well?

markreesmoj (approved):
👍

PR #68, opened 2024-01-11
Cpr 126

vb-moj:


markreesmoj (approved):
👍 although it would be nice to see a test proving that this fixes the problem - I appreciate that it is possibly difficult to do if this is a race condition

vb-moj:
> 👍 although it would be nice to see a test proving that this fixes the problem - I appreciate that it is possibly difficult to do if this is a race condition

We did update our IT but couldn't replicate the issue, but the tests now verifies the expected behaviour.

PR #70, opened 2024-01-11
CPR-128 Refactor  PNC validator to allow for PNC without forward slashes in

cairnsp:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/validate/PNCIdValidator.kt
```diff
@@ -14,28 +14,29 @@ class PNCIdValidator {
   }
 
   private fun hasValidCheckDigit(pncIdentifier: String): Boolean {
-    val checkDigit = pncIdentifier.last().uppercaseChar()
-    val modulus = calculateModulusOfSerialNumber(pncIdentifier)
+    val pncMinusSlash = pncIdentifier.replace("/", "") // removing forward slash
```
markreesmoj, line 17: I think the variable is well named enough not to need the comment

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/validate/PNCIdValidator.kt
```diff
@@ -14,28 +14,29 @@ class PNCIdValidator {
...
+    val modulus = calculateModulusOfSerialNumber(pncMinusSlash)
     val calculatedCheckDigit = convertNumberToLetterInAlphabet(modulus.toInt())
     return checkDigit == calculatedCheckDigit
   }
 
   private fun isValidFormat(pncIdentifier: String): Boolean = pncIdentifier.matches(Regex(PNC_REGEX))
 
   private fun calculateModulusOfSerialNumber(pncIdentifier: String): BigInteger {
-    // Starting format is : YYYY/NNNNNNND
+    // Starting format is : YYYY/NNNNNNND OR YYYYNNNNNNND
```
markreesmoj, line 27: Is this true? The slash has been removed by this point hasn't it?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/validate/PNCIdValidator.kt
```diff
@@ -14,28 +14,29 @@ class PNCIdValidator {
...
     val year = extractTwoDigitYearPart(pncIdentifier) // YY
-    val serialNumber = extractSerialNumberFromId(pncIdentifier) // NNNNNNN
+    val serialNumber = extractSerialNumberFromId(pncIdentifier, year) // NNNNNNN
```
markreesmoj, line 29: If I am reading this correctly, we don't actually need year and serialNumber except to combine them

`pncIdentifier.substring(2, 9)`

Looks like it might do the job

Although it may be less clear than what is currently in place, which is very easy to follow

vb-moj (approved):

markreesmoj (approved):
👍

PR #83, opened 2024-01-15
CPR-129 Prohibit defendant record creation where multiple records exist

cairnsp:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -45,7 +45,7 @@ class CourtCaseEventsService(
         } else if (matchesExistingRecordPartially(defendants, person)) {
           log.info("Partially matching CPR record exists for defendant - no further processing will occur")
           telemetryService.trackEvent(TelemetryEventType.NEW_CASE_PARTIAL_MATCH, extractMatchingFields(defendants[0], person))
-        } else {
+        } else if (defendants.isEmpty()) {
```
vb-moj, line 48: I think this will prevent the retry of any of the search failures, no?

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsServiceTest.kt
```diff
@@ -169,4 +170,29 @@ class CourtCaseEventsServiceTest {
     verify(offenderService).processAssociatedOffenders(personEntity, person)
     verify(prisonerService).processAssociatedPrisoners(personEntity, person)
   }
+
+  @Test
+  fun `should not create new defendant when mulitple existing records are found`() {
```
markreesmoj, line 175: minor: sp in multiple

markreesmoj (approved):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
vb-moj: Agree, but I still believe we should be retrying any of the search failures so will discuss that with them.

PR #85, opened 2024-01-15
REFACTOR: simplify dependencies of CourtCaseService

markreesmoj:
I was playing around with the CourtCaseService and this change seemed small and a slight improvement

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
@@ -11,13 +11,13 @@ import uk.gov.justice.digital.hmpps.personrecord.validate.PNCIdValidator
 @Service
 class CourtCaseEventsService(
   private val telemetryService: TelemetryService,
-  private val pncIdValidator: PNCIdValidator,
   private val defendantRepository: DefendantRepository,
   private val personRecordService: PersonRecordService,
   private val offenderService: OffenderService,
   private val prisonerService: PrisonerService,
 ) {
 
+  private val pncIdValidator: PNCIdValidator = PNCIdValidator()
   companion object {
     private val log = LoggerFactory.getLogger(this::class.java)
   }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -11,13 +11,13 @@ import uk.gov.justice.digital.hmpps.personrecord.validate.PNCIdValidator
...
 class CourtCaseEventsService(
   private val telemetryService: TelemetryService,
-  private val pncIdValidator: PNCIdValidator,
   private val defendantRepository: DefendantRepository,
   private val personRecordService: PersonRecordService,
   private val offenderService: OffenderService,
   private val prisonerService: PrisonerService,
 ) {
 
+  private val pncIdValidator: PNCIdValidator = PNCIdValidator()
```
markreesmoj, line 20: This doesn't have any Spring dependencies so can just be initialised here

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsServiceTest.kt
```diff
@@ -35,9 +33,6 @@ class CourtCaseEventsServiceTest {
   @Mock
   lateinit var offenderService: OffenderService
 
-  @Mock
-  lateinit var pncIdValidator: PNCIdValidator
-
```
markreesmoj, line 40: No need to mock as the CourtCaseService creates its own instance of PNCIdValidator

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsServiceTest.kt
```diff
@@ -74,11 +69,11 @@ class CourtCaseEventsServiceTest {
   @Test
   fun `should call telemetry service when exact match found for Court Case Event`() {
     // Given
-    val pncNumber = "PNC12345"
+    val pncNumber = "20030011985X"
```
markreesmoj, line 72: Had to do this as real validation is now happening - I borrowed this ID from the PNCIdValidatorTest

PR #86, opened 2024-01-16
CPR-129 Semantic fix to prevent multiple record creation

cairnsp:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -37,17 +37,17 @@ class CourtCaseEventsService(
         log.info("Invalid PNC encountered in court case event - no further processing will occur")
         telemetryService.trackEvent(TelemetryEventType.NEW_CASE_INVALID_PNC, mapOf("PNC" to pnc))
       } else {
-        val defendants = defendantRepository.findAllByPncNumber(pnc)
+        val defendants = personRepository.findByDefendantsPncNumber(pnc)?.defendants.orEmpty()
```
vb-moj, line 40: Agree, this is the right way to go but I still believe this won't change the issue we are having about duplication.

vb-moj (approved):

PR #96, opened 2024-01-19
Double count empty string pnc

markreesmoj:
At the moment, an empty string PNC is treated as both missing and invalid (see the test added in this PR).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
@@ -56,7 +56,7 @@ class CourtCaseEventsService(
             offenderService.processAssociatedOffenders(it, person)
             prisonerService.processAssociatedPrisoners(it, person)
           }
-        }
+        } // what if defendants is not empty?
       }
     }
   }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
@@ -28,27 +28,25 @@ class CourtCaseEventsService(
   fun processPersonFromCourtCaseEvent(person: Person) {
     log.debug("Entered processPersonFromCourtCaseEvent")
     val pnc: String? = person.otherIdentifiers?.pncNumber
-    // Validate PNC
-    if (!pncIdValidator.isValid(pnc)) {
-      return
-    }
-    val defendants = personRepository.findByDefendantsPncNumber(pnc!!)?.defendants.orEmpty()
+    if (pncIdValidator.isValid(pnc)) {
+      val defendants = personRepository.findByDefendantsPncNumber(pnc!!)?.defendants.orEmpty()
 
-    if (matchesExistingRecordExactly(defendants, person)) {
-      log.info("Exactly matching Person record exists with defendant - no further processing will occur")
-      telemetryService.trackEvent(TelemetryEventType.NEW_CASE_EXACT_MATCH, mapOf("PNC" to pnc, "CRN" to defendants[0].crn, "UUID" to person.personId.toString()))
-    } else if (matchesExistingRecordPartially(defendants, person)) {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -19,45 +19,33 @@ class CourtCaseEventsService(
...
+    if (pncIdValidator.isValid(pnc)) {
+      val defendants = personRepository.findByDefendantsPncNumber(pnc!!)?.defendants.orEmpty()
 
-    // Ensure PNC is present
-    if (person.otherIdentifiers?.pncNumber.isNullOrEmpty()) {
-      log.info("PNC not present in court case event - no further processing will occur")
-      telemetryService.trackEvent(TelemetryEventType.NEW_CASE_MISSING_PNC, emptyMap())
-    }
-
-    person.otherIdentifiers?.pncNumber?.let { pnc ->
```
markreesmoj, line 37: This allows an empty string into the method

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/validate/PNCIdValidator.kt
```diff
@@ -1,14 +1,25 @@
...
 const val PNC_REGEX = "\\d{4}(/?)\\d{7}[A-Z]{1}\$"
 
-class PNCIdValidator {
+class PNCIdValidator(private val telemetryService: TelemetryService) {
 
-  fun isValid(pncIdentifier: String): Boolean {
-    return isValidFormat(pncIdentifier) && hasValidCheckDigit(pncIdentifier)
+  fun isValid(pncIdentifier: String?): Boolean {
+    if (pncIdentifier.isNullOrEmpty()) {
+      telemetryService.trackEvent(TelemetryEventType.NEW_CASE_MISSING_PNC, emptyMap())
```
vb-moj, line 14: This implies the case is missing the PNC and may not be correct for other services. If we want the validator should send the event then the event name should be generic or move to the calling services.

vb-moj (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/validate/PNCIdValidator.kt
markreesmoj: Fair point, I didn't want to change the existing functionality. Perhaps something like `TelemetryEventType.MISSING_PNC`?
Would need to run it past a BA

markreesmoj: This assumes that we are going to get and validate PNCs from other sources

markreesmoj: I suggest that we hold off renaming the event until we have another source of PNCs - at the moment it is correct

markreesmoj: I have renamed the event

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -19,45 +19,33 @@ class CourtCaseEventsService(
...
+      } else if (defendants.isEmpty()) {
+        log.debug("No existing matching Person record exists - creating new person and defendant with pnc $pnc")
+        val personRecord = personRecordService.createNewPersonAndDefendant(person)
+        telemetryService.trackEvent(TelemetryEventType.NEW_CASE_PERSON_CREATED, mapOf("UUID" to personRecord.personId.toString(), "PNC" to pnc))
+        personRecord.let {
+          offenderService.processAssociatedOffenders(it, person)
+          prisonerService.processAssociatedPrisoners(it, person)
         }
-      }
+      } // what if defendants is not empty?
```
vb-moj, line 48: Is this still needed?

vb-moj (approved):

PR #102, opened 2024-01-22
CPR-144 Implement PNC ID mismatch telemetry event

cairnsp:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PrisonerService.kt
```diff
@@ -27,26 +27,36 @@ class PrisonerService(
...
       val prisoners = client.findPossibleMatches(PossibleMatchCriteria.from(person))
       log.debug("Number of prisoners returned from Nomis for PNC ${person.otherIdentifiers?.pncNumber} = ${prisoners?.size ?: 0}")
-      prisoners?.let {
-        if (matchesExistingPrisonerExactly(it, person)) {
+      prisoners?.let { prisonerList ->
+        if (pncIdentifierDoesNotMatch(prisonerList, person)) {
+          log.debug("Nomis PNC Id does not match that of person")
+          telemetryService.trackEvent(
+            TelemetryEventType.NOMIS_PNC_DOES_NOT_MATCH,
+            mapOf(
```
vb-moj, line 35: I think we should also create a new person for non-matching PNC, no? 🤔

vb-moj (changes requested):
Added a comment for some clarification.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PrisonerService.kt
markreesmoj: Maybe, in order to match the behaviour of the HMCTS listener, although based on recent discussions it seems PNC may not be suitable as a unique identifier

vb-moj (approved):

PR #108, opened 2024-01-24
CPR-148 Add PNCIdentifier class to standardize PNC id format

cairnsp:
A new class, PNCIdentifier, has been added to handle the conversion of PNC IDs to a standardized canonical format. This standardization is useful when comparing PNC ids across different person related entites/pojos.
 
PNCIdentifier provides two main methods: 

-  toCanonicalForm :  for converting a PNC ID to the canonical format
-  areEquivalent:  for comparing two PNC IDs 
 
All references to PNC id equality checks and conversions have been updated to use these new methods.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/ProbationOffenderSearchClientIntTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/ProbationOffenderSearchClientIntTest.kt
@@ -42,7 +42,7 @@ class ProbationOffenderSearchClientIntTest : IntegrationTestBase() {
   @Test
   fun `should return bad request for insufficient search parameters`() {
     // Given
-    val searchDto = SearchDto()
+    val searchDto = SearchDto(includeAliases = null)
 
     // When
     val exception = assertThrows(FeignException.BadRequest::class.java) {
--- a/src/test/resources/logback.xml
+++ b/src/test/resources/logback.xml
@@ -38,6 +38,10 @@
         <appender-ref ref="CONSOLE"/>
     </logger>
 
+    <logger name="com.github.tomakehurst.wiremock" additivity="false" level="DEBUG">
+        <appender-ref ref="CONSOLE"/>
+    </logger>
+
     <logger name="io.netty" additivity="false" level="ERROR">
         <appender-ref ref="CONSOLE"/>
     </logger>
--- a/src/test/resources/mappings/offender-search/POST_search_bad_request.json
+++ b/src/test/resources/mappings/offender-search/POST_search_bad_request.json
@@ -1,4 +1,5 @@
 {
+  "priority": 3,
   "request": {
     "method": "POST",
     "url": "/search",
@@ -12,7 +13,7 @@
           "croNumber" : null,
           "crn" : null,
           "nomsNumber" : null,
-          "includeAliases" : false
+          "includeAliases" : null
         }
       }
     ]
--- a/src/test/resources/mappings/prisoner-search/POST_search_default_result.json
+++ b/src/test/resources/mappings/prisoner-search/POST_search_default_result.json
@@ -1,7 +1,7 @@
 {
   "request": {
     "method": "POST",
-    "url": "/search"
+    "url": "/prisoner-search/possible-matches"
   },
   "response": {
     "status": 200,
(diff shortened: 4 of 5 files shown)


markreesmoj:
Just a general thought - might it be an idea to make PncNumber into a type and have some/all of the associated methods on it?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -27,25 +27,38 @@ class CourtCaseEventsService(
   @Transactional(isolation = Isolation.SERIALIZABLE)
   fun processPersonFromCourtCaseEvent(person: Person) {
     log.debug("Entered processPersonFromCourtCaseEvent")
-    val pnc: String? = person.otherIdentifiers?.pncNumber
-    if (pncIdValidator.isValid(pnc)) {
-      val defendants = personRepository.findByDefendantsPncNumber(pnc!!)?.defendants.orEmpty()
+    if (person.otherIdentifiers?.pncNumber?.isNotEmpty() == true) {
+      val pncId = person.otherIdentifiers.pncNumber
```
markreesmoj, line 31: wouldn't we want to get the canonical form here?
presumably we are going to have to empty the database when this is merged?

markreesmoj:
https://github.com/ministryofjustice/hmpps-person-record/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processor/CourtCaseEventsProcessor.kt#L53 is a pnc which comes from a message and is not sanitised

Indeed, it seems like there are quite a few places where we are not sanitising the PNC number in the code - it is hard to be sure whether this is a problem or not. For example, compare

https://github.com/ministryofjustice/hmpps-person-record/blob/950789e4bb992e9ba428696a39d36e357470f541/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/Person.kt#L83
and 
https://github.com/ministryofjustice/hmpps-person-record/blob/950789e4bb992e9ba428696a39d36e357470f541/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/Person.kt#L73

As far as I can tell, both are designed to construct a Person from a JSON structure deserialised from a message.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -84,7 +99,7 @@ class CourtCaseEventsService(
       .and(defendants.size == 1)
       .and(
         defendants.any {
-          it.pncNumber.equals(person.otherIdentifiers?.pncNumber) &&
+          person.otherIdentifiers?.pncIdentifier?.isEquivalentTo(PNCIdentifier(it.pncNumber)) == true &&
```
markreesmoj, line 102: is `== true` necessary? isEquivalentTo will return a Boolean

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/validate/PNCIdentifier.kt
```diff
@@ -0,0 +1,64 @@
+package uk.gov.justice.digital.hmpps.personrecord.validate
+
+import org.slf4j.LoggerFactory
+import java.time.LocalDate
+
+const val LONG_PNC_ID_LENGTH = 10
+
+data class PNCIdentifier(private val inputPncId: String? = null) {
```
markreesmoj, line 8: Love a data class, I have missed them for the past year. Java Records look like they might be good too

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/validate/PNCIdentifier.kt
```diff
@@ -0,0 +1,64 @@
...
+  val pncId: String?
+    get() = toCanonicalForm(inputPncId)
+
+  fun isEquivalentTo(otherPncId: PNCIdentifier?): Boolean {
```
markreesmoj, line 12: I wonder if this could be the `equals` method now? Not sure it is a good idea but worth thinking about

markreesmoj (approved):
A big step in the right direction - this shows just how wedded we are to PNC at the moment!

markreesmoj (approved):
👍

PR #120, opened 2024-01-29
CPR-145 test updates

cairnsp:


vb-moj (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/CourtCaseEventsListenerIntTest.kt
```diff
@@ -167,11 +166,19 @@ class CourtCaseEventsListenerIntTest : IntegrationTestBase() {
...
+        eq(TelemetryEventType.INVALID_PNC),
+        check {
+          assertThat(it["PNC]"]).isNullOrEmpty()
+        },
+      )
+    }
   }
 
   @Test
-  @Disabled("Works locally - disabled until reason for failure on circleCi has been established")
```
markreesmoj, line 173: Great to see this has gone - we should also enable or delete all the other disabled tests.

PR #122, opened 2024-01-29
Persist more offender data from nDelius search response.

vb-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/OffenderEntity.kt
```diff
@@ -55,6 +55,10 @@ class OffenderEntity(
       return person.otherIdentifiers?.crn?.let {
         val offenderEntity = OffenderEntity(
           crn = it,
+          pncNumber = person.otherIdentifiers.pncIdentifier?.pncId,
```
markreesmoj, line 58: 👍 although I think we ought to consider at some point persisting the PNCIdentifier type itself here

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/Person.kt
```diff
@@ -55,7 +55,10 @@ data class Person(
 
     fun from(offenderDetail: OffenderDetail): Person {
       return Person(
-        otherIdentifiers = OtherIdentifiers(crn = offenderDetail.otherIds.crn),
+        givenName = offenderDetail.firstName,
+        familyName = offenderDetail.surname,
+        dateOfBirth = offenderDetail.dateOfBirth,
+        otherIdentifiers = OtherIdentifiers(crn = offenderDetail.otherIds.crn, pncIdentifier = PNCIdentifier(offenderDetail.otherIds.pncNumber)),
```
markreesmoj, line 61: Similarly 👍 but we should think about naming - we have PncId, PncIdentifer, PncNumber just in this small code sample - they are all the same thing

markreesmoj (approved):
👍

PR #124, opened 2024-01-29
CPR-162

markreesmoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/CourtCaseEventsListenerIntTest.kt
```diff
@@ -68,12 +67,9 @@ class CourtCaseEventsListenerIntTest : IntegrationTestBase() {
       .build()
 
     // when
-    val publishResponse = courtCaseEventsTopic?.snsClient?.publish(publishRequest)?.get()
+    courtCaseEventsTopic?.snsClient?.publish(publishRequest)?.get()
 
     // then
-    assertThat(publishResponse?.sdkHttpResponse()?.isSuccessful).isTrue()
-    assertThat(publishResponse?.messageId()).isNotNull()
-
```
markreesmoj, line 76: These assertions don't seem necessary - the following `await` will block and time out if the publish has failed for any reason

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/CourtCaseEventsListenerIntTest.kt
```diff
@@ -166,15 +156,6 @@ class CourtCaseEventsListenerIntTest : IntegrationTestBase() {
...
     }
-
-    await untilAsserted {
-      verify(telemetryService).trackEvent(
-        eq(TelemetryEventType.INVALID_PNC),
-        check {
-          assertThat(it["PNC]"]).isNullOrEmpty()
-        },
-      )
-    }
```
markreesmoj, line 177: I don't understand why this happens really but it does not seem to be desirable behaviour and the application does not actually do this when deployed as far as I can see.

perhaps it is something to do with this?
https://github.com/spring-projects/spring-data-rest/issues/1440

At all events I don't think we should assert that it happens

PR #132, opened 2024-01-30
CPR-113 Handle multiple delius matches

cairnsp:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/OffenderService.kt
```diff
@@ -18,52 +18,67 @@ class OffenderService(
...
+
       if (offenderDetails.isNullOrEmpty()) {
-        log.debug("No Delius matching records exist")
-        telemetryService.trackEvent(
-          TelemetryEventType.DELIUS_NO_MATCH_FOUND,
-          mapOf("UUID" to personEntity.personId.toString(), "PNC" to person.otherIdentifiers?.pncIdentifier?.pncId),
-        )
+        logAndTrackEvent(NO_RECORDS_MESSAGE, TelemetryEventType.DELIUS_NO_MATCH_FOUND, personEntity, person)
+      } else {
+        when {
```
markreesmoj, line 37: Nice - like this way of using `when`

What happens if none of these conditions are met?

markreesmoj (approved):
👍

vb-moj (approved):

PR #138, opened 2024-02-01
Cpr 145 catch cannot acquire lock exception

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -52,21 +52,23 @@ class CourtCaseEventsService(
...
           personRecord.let {
             offenderService.processAssociatedOffenders(it, person)
             prisonerService.processAssociatedPrisoners(it, person)
           }
+
+          telemetryService.trackEvent(
+            TelemetryEventType.NEW_CASE_PERSON_CREATED,
+            mapOf("UUID" to personRecord.personId.toString(), "PNC" to pncId),
+          )
+          log.error("NEW CASE $personRecord.personId.toString() PNC $pncId")
```
markreesmoj, line 71: Just put this here so it is easier to spot when the test fails

vb-moj (approved):

PR #140, opened 2024-02-01
CPR-145 remove JPA from auto setup for tests

markreesmoj:
The jpa config does not appear to have any effect. Flyway is always used (I have checked this on main)


src/test/resources/sql/after-test.sql
```diff
@@ -1 +1,4 @@
-TRUNCATE TABLE person CASCADE;
\ No newline at end of file
+TRUNCATE TABLE person CASCADE;
+TRUNCATE TABLE offender CASCADE;
+TRUNCATE TABLE defendant CASCADE;
+TRUNCATE TABLE prisoner CASCADE;
```
markreesmoj, line 4: I think we need to do this because the JPA setting have been removed

src/test/resources/application-test.yml
```diff
@@ -28,14 +28,6 @@ feature:
...
     enable-nomis-search: true
 
-spring:
-  jpa:
-    properties:
-      hibernate:
-        hbm2ddl:
-          auto: update
-    show-sql: false
-
```
vb-moj, line 38: Why do we need to remove this?

vb-moj: This is in the test so not much to bother with, it is just that we need to manually do things like the one above.

markreesmoj:
This seems to expose the problem with a DB structure which does not match what is defined in code. See https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-person-record/842/workflows/1a473957-9ed4-45eb-8463-2b3c209faab9/jobs/2488

vb-moj (approved):

PR #143, opened 2024-02-01
CPR 149 ensure pnc ids are valid across tests

cairnsp:
- Tidied up PNC ids in SQL scripts and test data
- Added custom JPA converter for PNCIdentifier
- Refactored Offender, Defendant and Prisoner entities to use PNCIdentifier
- Refactor of associated repository methods, and tests


vb-moj (commented):
LGTM except the  verious names used here. PNCIdentifier, PNCIdvalidator, pncNumber. Happy to have a chat on this.

markreesmoj (approved):
This is a massive improvement, I love this work. The constants are good, increased encapsulation of the PncIdentifier class instead of reaching inside it to get the Pnc as a string, refactoring of the validation all looks really good.

I wonder if this is a good point to truncate the live database - is there anything currently in there which will cause problems with this change? Records we may be unable to update perhaps as we cannot validate the PNC we have read from the DB?

PR #156, opened 2024-02-07
CPR-141 do not allow access to non-canonical format of PNC

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -34,7 +34,7 @@ class CourtCaseEventsService(
     person.otherIdentifiers?.pncIdentifier?.let { pncIdentifier ->
       val pncId = pncIdentifier.pncId
       if (pncIdentifier.isValid()) {
-        telemetryService.trackEvent(TelemetryEventType.VALID_PNC, mapOf("PNC" to pncIdentifier.pncId, "inputPNC" to pncIdentifier.inputPnc))
+        telemetryService.trackEvent(TelemetryEventType.VALID_PNC, mapOf("PNC" to pncIdentifier.toString()))
```
markreesmoj, line 37: We were logging inputPNC as part of the work to determine if the validator was working correctly. I don't think it's necessary any more - if it turns out that it is, we can do it when canonicalising the PNC

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/validate/PNCIdentifier.kt
```diff
@@ -5,12 +5,12 @@ import java.time.LocalDate
 
 const val LONG_PNC_ID_LENGTH = 10
 
-data class PNCIdentifier(private val inputPncId: String? = null) {
-  val inputPnc: String?
-    get() = inputPncId
+class PNCIdentifier(private val inputPncId: String? = null) {
```
markreesmoj, line 8: We don't seem to be using the capabilities of the data class here - we provide an alternative getter - so I have removed it

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/validate/PNCIdentifier.kt
```diff
@@ -5,12 +5,12 @@ import java.time.LocalDate
...
+  private val storedPncId: String? = inputPncId?.uppercase()
+  // always use canonical format when storing and comparing
 
   val pncId: String?
-    get() = toCanonicalForm(inputPncId)
+    get() = toCanonicalForm(storedPncId)
```
markreesmoj, line 13: I wonder if we might be better with the factory pattern to create a PNC and do all the necessary logging?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/validate/PNCIdentifier.kt
```diff
@@ -5,12 +5,12 @@ import java.time.LocalDate
 
 const val LONG_PNC_ID_LENGTH = 10
 
-data class PNCIdentifier(private val inputPncId: String? = null) {
-  val inputPnc: String?
-    get() = inputPncId
+class PNCIdentifier(inputPncId: String? = null) {
+  private val storedPncId: String = inputPncId?.uppercase() ?: ""
```
markreesmoj, line 9: I don't think we need to differentiate between null and empty PNCs

vb-moj (commented):
LGTM.

vb-moj (approved):

PR #158, opened 2024-02-08
Cpr 154 defendant matcher

vb-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/matcher/DefendantMatcher.kt
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  fun extractMatchingFields(defendant: DefendantEntity): Map<String, String?> =
+    mapOf(
+      "Surname" to if (defendant.surname.equals(person.familyName)) defendant.surname else null,
+      "Forename" to if (defendant.forenameOne.equals(person.givenName)) defendant.forenameOne else null,
+      "Date of birth" to if (defendant.dateOfBirth?.equals(person.dateOfBirth) == true) defendant.dateOfBirth.toString() else null,
+    ).filterValues { it != null }
+
+  fun hasNothingMatch(): Boolean {
```
markreesmoj, line 29: I don't really understand this method - the other matchers don't have it. Does it belong here?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/matcher/Matcher.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.personrecord.service.matcher
+
+import uk.gov.justice.digital.hmpps.personrecord.model.Person
+
+abstract class Matcher<T>(val items: List<T>?, protected val person: Person) {
```
markreesmoj, line 5: This could be an interface I think - it doesn't have any state

markreesmoj: I do really like this change by the way

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/matcher/DefendantMatcher.kt
vb-moj: Yes, only the defendant matcher needs this.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/matcher/Matcher.kt
vb-moj: We could use an interface but I prefer this approach as this is simpler.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/matcher/DefendantMatcher.kt
vb-moj: I also thought about the need for this method, the only reason we added that empty check before was during the initial release when we had numerous duplications. will try to move this logic to the else block and check.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -28,78 +33,55 @@ class CourtCaseEventsService(
...
-            NEW_CASE_PERSON_CREATED,
-            mapOf("UUID" to personRecord.personId.toString(), "PNC" to pncId),
-          )
-        } // what if defendants is not empty?
+        when {
+          defendantMatcher.isExactMatch() -> exactMatchFound(defendantMatcher, person, pncId)
+          defendantMatcher.isPartialMatch() -> partialMatchFound(defendantMatcher)
+          else -> {
+            createNewPersonRecordAndProcess(person, pncId)
+          }
```
vb-moj, line 51: Hope this works without causing any issues. 🤞

markreesmoj: I think it will be fine - the way the old code was written led you to believe that there was a possibility that non-matching defendants had been found by `personRepository.findByDefendantsPncNumber(PNCIdentifier(pncId))?.defendants.orEmpty()`. 
I don't think that is possible.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/matcher/OffenderMatcher.kt
```diff
@@ -4,40 +4,20 @@ import uk.gov.justice.digital.hmpps.personrecord.client.model.OffenderDetail
...
-    return !offenderDetails.isNullOrEmpty() && offenderDetails.size > 1 && offenderDetails.all { isMatchingOffender(person, it) }
-  }
-
+class OffenderMatcher(offenderDetails: List<OffenderDetail>?, person: Person) :
+  Matcher<OffenderDetail>(offenderDetails, person) {
   fun isPncDoesNotMatch(): Boolean {
-    return !offenderDetails.isNullOrEmpty() && offenderDetails.none {
-      person.otherIdentifiers?.pncIdentifier == PNCIdentifier(it.otherIds.pncNumber)
+    return !items.isNullOrEmpty() && items.none {
+      person.otherIdentifiers?.pncIdentifier == PNCIdentifier(it.otherIds.pncNumber.toString())
```
markreesmoj, line 11: `it.otherIds.pncNumber` is already a `String?` - no need for the toString() call

markreesmoj: I think actually making IDs.pncNumber into a PNCIdentifier is probably the right way to do things, but that is not necessary for this PR :-)

markreesmoj (approved):

PR #159, opened 2024-02-08
CPR-141 use a factory to create PNCIdentifier

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/validate/PNCIdentifier.kt
```diff
@@ -5,11 +5,21 @@ import java.time.LocalDate
 
 const val LONG_PNC_ID_LENGTH = 10
 
-class PNCIdentifier(inputPncId: String? = null) {
+interface PNCIdentifier {
+
+  val pncId: String
+  companion object {
+    fun create(inputPncId: String? = null) = ValidPNCIdentifier(inputPncId)
```
markreesmoj, line 12: will soon do the validation in here and create the correct type of PNCIdentifier

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/controller/PersonControllerIntTest.kt
```diff
@@ -174,7 +174,6 @@ class PersonControllerIntTest() : IntegrationTestBase() {
       familyName = "Jones",
       dateOfBirth = LocalDate.of(1968, 8, 15),
       otherIdentifiers = OtherIdentifiers(
-        pncIdentifier = PNCIdentifier("1976/1234Z"),
```
markreesmoj, line 177: This test was a new one on me. I don't think we are anywhere near supporting this functionality and based on the architecture diagram, it may not even be in this application. I think it might be an idea to remove it for the moment but have bodged it for now. If you supply a PNC it won't work.

vb-moj (commented):
Just one comment.

PR #162, opened 2024-02-09
CPR-141 custom repository

markreesmoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonRepositoryCustomImplIntTest.kt
```diff
@@ -33,25 +29,11 @@ class PersonRepositoryCustomImplIntTest : IntegrationTestBase() {
...
-      forenameThree = "bob",
-    )
-
-    // When
-    val results = personRepository.searchByRequestParameters(searchRequest)
-
-    // Then
-    assertThat(results).hasSize(1)
-    assertThat(results[0].personId).isEqualTo(UUID.fromString("eed4a9a4-d853-11ed-afa1-0242ac120002"))
-  }
```
markreesmoj, line 50: creation of a person with middle names is unsupported at the moment

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/integration/IntegrationTestBase.kt
```diff
@@ -127,4 +134,16 @@ abstract class IntegrationTestBase {
...
+
+    val newPersonEntity = PersonEntity(personId = personId)
+    newPersonEntity.createdBy = "test"
+    newPersonEntity.lastUpdatedBy = "test"
+    val newDefendantEntity = DefendantEntity.from(person)
+    newDefendantEntity.person = newPersonEntity
+    newPersonEntity.defendants.add(newDefendantEntity)
+    personRepository.saveAndFlush(newPersonEntity)
+  }
 }
```
vb-moj, line 149: This class is getting bigger. I think we need to start thinking about moving the common data to separate files.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonRepositoryCustomImplIntTest.kt
vb-moj: The ticket I just picked up mentioned this so better keep it.

markreesmoj: Is the functionality well-defined? It does not work at the moment, in the sense that any data created by our app will only have first name (given name) and family name, so this search can't succeed. Moreover, the existing `from` method in PersonSearchRequest does not allow you to create a search on middle name, and that is what is currently used in production. 

So this test will only work on data inserted directly into the database - something which does not happen when consuming messages.

By all means we can add the test back in once it is testing something which is in use in production code - test driving that implementation perhaps.

vb-moj (approved):

PR #164, opened 2024-02-09
Cpr 141 multiple implementations

markreesmoj:
PNCIdentifer must now be valid in order to be written into the database when coming from nDelius (or any other source)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -49,9 +47,9 @@ class CourtCaseEventsService(
             createNewPersonRecordAndProcess(person, pncId)
           }
         }
-      } else {
-        trackEvent(INVALID_PNC, mapOf("PNC" to pncIdentifier.toString()))
       }
+      is InvalidPNCIdentifier -> trackEvent(INVALID_PNC, mapOf("PNC" to pncIdentifier.invalidValue()))
+      else -> trackEvent(MISSING_PNC, emptyMap())
     }
```
vb-moj, line 53: A small suggestion if you are interested! we could extract further so that the code will looks like this

`    when (val pncIdentifier = person.otherIdentifiers?.pncIdentifier) {
      is ValidPNCIdentifier -> processValidPNCRecord(pncIdentifier, person)
      is InvalidPNCIdentifier -> processInvalidPNCRecord(pncIdentifier)
      else -> trackEvent(MISSING_PNC, emptyMap())
    }`

vb-moj (commented):
One minor comment.

vb-moj (approved):
One minor comment.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsServiceTest.kt
```diff
@@ -189,25 +188,4 @@ class CourtCaseEventsServiceTest {
...
-    )
-
-    // When
-    courtCaseEventsService.processPersonFromCourtCaseEvent(person)
-
-    // Then
-    verifyNoInteractions(personRecordService)
-    verifyNoInteractions(offenderService)
-    verifyNoInteractions(prisonerService)
-  }
```
markreesmoj, line 212: This doesn't test what it says it does. It only passes because the PNC is in an invalid format, and in my view is covered by the earlier test `should call telemetry service when exact match found for Court Case Event`

PR #165, opened 2024-02-14
CPR-153 Added more fields from HMCTS.

vb-moj:
Updated required data models.

Updated defendant entity.


vb-moj:
We no longer need this branch.

PR #166, opened 2024-02-14
CPR-159 Review / Restructure JPA Entities

vb-moj:
**PLEASE DO NOT MERGE.** 

Steps to take before the deployment.

1. Stop the message processing
2. Truncate all tables
3. Deploy CPR-153 & CPR-159
4. Restart message processing


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/AddressEntity.kt
```diff
@@ -0,0 +1,54 @@
...
+
+  @Column(name = "address_line_three")
+  val addressLineThree: String? = null,
+
+  @Column(name = "address_line_four")
+  val addressLineFour: String? = null,
+
+  @Column(name = "address_line_five")
+  val addressLineFive: String? = null,
+
```
markreesmoj, line 35: Is there anything to be said for storing address as just one string with newlines? It might facilitate comparisons with other systems. 5 lines of address is an arbitrary number anyway, what if one has six?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/AddressEntity.kt
```diff
@@ -0,0 +1,54 @@
...
+  @Column(name = "postcode")
+  val postcode: String? = null,
+) : BaseAuditedEntity() {
```
markreesmoj, line 38: This should definitely be split out from the address though

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/ContactEntity.kt
```diff
@@ -0,0 +1,42 @@
...
+  @Column(name = "mobile")
+  val mobile: String? = null,
+
+  @Column(name = "primary_email")
+  val primaryEmail: String? = null,
+
+) : BaseAuditedEntity() {
+  companion object {
+    fun from(person: Person): ContactEntity {
+      val addressEntity = ContactEntity()
```
markreesmoj, line 36: Minor but variable should perhaps be named contactEntity

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/DefendantAliasEntity.kt
```diff
@@ -0,0 +1,38 @@
...
+  var defendant: DefendantEntity? = null,
+
+  @Column(name = "first_name")
+  val firstName: String? = null,
+
+  @Column(name = "middle_name")
+  val middleName: String? = null,
+
+  @Column(name = "last_name")
+  val lastName: String? = null,
```
markreesmoj, line 37: As per address, is there anything to be said for simply having a "name" field? What if their name is Cristiano Ronaldo dos Santos Aveiro for example?

markreesmoj: But if we do decide to have it split, let us be consistent. We have a `surname` on defendant and `lastName` on defendant_alias

src/main/resources/db/migration/V7__add_address_contact_defendant_alias_tables.sql
```diff
@@ -0,0 +1,134 @@
...
+    postcode                                TEXT        DEFAULT NULL,
+    created_by                              TEXT        NOT NULL,
+    created_date                            TIMESTAMP   NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    last_updated_by                         TEXT        NOT NULL,
+    last_updated_date                       TIMESTAMP   NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    version                                 int4        NOT NULL DEFAULT 0
+
+);
+
+DROP TABLE IF EXISTS address_aud;
```
markreesmoj, line 22: I think we should drop this auditing and use hmpps_audit if we do have audit requirements (which at the moment for me we do not) - Phil kept complaining about it and it seems to cause us problems

src/main/resources/db/migration/V7__add_address_contact_defendant_alias_tables.sql
```diff
@@ -0,0 +1,134 @@
...
+    DROP COLUMN address_line_five,
+    DROP COLUMN postcode;
+
+
+ALTER table defendant
+    Add column fk_address_id bigInt NULL;
+
+alter table if exists defendant_aud add column fk_address_id bigInt;
+
+ALTER TABLE defendant ADD CONSTRAINT fk_address FOREIGN KEY (fk_address_id) REFERENCES address;
```
markreesmoj, line 125: Does this make address a mandatory field(s)? This is not the case at the moment

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/AddressEntity.kt
vb-moj: we go with what we have currently and update in the future.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/DefendantAliasEntity.kt
vb-moj: Raised this last week as well, We have decided to have a chat and decide later on.

src/main/resources/db/migration/V7__add_address_contact_defendant_alias_tables.sql
vb-moj: No, it does not. It just enforces the referential integrity constraint that's all.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/DefendantAliasEntity.kt
vb-moj: I am going keep the surname on both tables. if they want to to change will change latter.

vb-moj:
Created a new branch so closed this.

PR #170, opened 2024-02-16
CPR-153 Fixed the error when checking the flag.

vb-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/OffenderDomainEventsListener.kt
```diff
@@ -33,7 +33,7 @@ class OffenderDomainEventsListener(
   ) {
     log.debug("Enter onDomainEvent")
     val sqsMessage = objectMapper.readValue<SQSMessage>(rawMessage)
-    featureFlag.isDeliusDomainEventSQSEnabled().let {
```
vb-moj, line 36: .let only check for null.

PR #171, opened 2024-02-19
Cpr 159 new tables

vb-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/DefendantEntity.kt
```diff
@@ -4,21 +4,22 @@ import jakarta.persistence.CascadeType
...
+import jakarta.persistence.FetchType
 import jakarta.persistence.GeneratedValue
 import jakarta.persistence.GenerationType
 import jakarta.persistence.Id
 import jakarta.persistence.JoinColumn
 import jakarta.persistence.ManyToOne
+import jakarta.persistence.OneToMany
+import jakarta.persistence.OneToOne
 import jakarta.persistence.Table
-import org.hibernate.envers.Audited
```
markreesmoj, line 13: I think we can get rid of this dependency from build.grade too

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/integration/IntegrationTestBase.kt
```diff
@@ -142,8 +142,6 @@ abstract class IntegrationTestBase {
     val person = Person(otherIdentifiers = OtherIdentifiers(pncIdentifier = pncIdentifier, crn = crn), givenName = givenName, familyName = familyName, dateOfBirth = dateOfBirth)
 
     val newPersonEntity = PersonEntity(personId = personId)
-    newPersonEntity.createdBy = "test"
-    newPersonEntity.lastUpdatedBy = "test"
```
markreesmoj, line 146: ❤️

src/main/resources/db/migration/V7__add_address_contact_defendant_alias_tables.sql
```diff
@@ -0,0 +1,94 @@
...
+--------- defendant_alias ------------------
+DROP TABLE IF EXISTS defendant_alias;
+
+CREATE TABLE IF NOT EXISTS defendant_alias
+(
+    id                                      SERIAL      PRIMARY KEY,
+    first_name                              TEXT        DEFAULT NULL,
+    middle_name                             TEXT        DEFAULT NULL,
+    surname                                 TEXT        DEFAULT NULL,
+    version                                 int4        NOT NULL DEFAULT 0,
```
markreesmoj, line 42: Do we have to handle this?
It looks like it used to be part of BaseAuditedEntity but as we have deleted that, maybe version is no longer required?

markreesmoj: Also, it might be worth switching message processing on in dev and perhaps also preprod (not sure how that fits with the security concerns but it is not live) before merging this, as that will show us if the changes actually work :-)

vb-moj: I am ok with this PR as making further changes is more complicated and has very little gain.

vb-moj: Sorry, my bad. we need this but forgot it was in the Baseaudit. Need to move the annotation to individual entities now. Thanks

markreesmoj (approved):
Provided that version works as expected this is OK

markreesmoj (approved):
👍

PR #173, opened 2024-02-19
Cpr 153

vb-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/PersonSearchRequest.kt
```diff
@@ -24,7 +22,7 @@ data class PersonSearchRequest(
       return PersonSearchRequest(
         pncNumber = person.otherIdentifiers?.pncIdentifier?.pncId,
         crn = person.otherIdentifiers?.crn,
-        forenameOne = person.givenName,
+        firstName = person.givenName,
```
markreesmoj, line 25: It's my impression that we don't need or use a lot of this functionality. However might it be worth adding in person.middleName here?

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/DefendantEntityTest.kt
```diff
@@ -17,6 +17,7 @@ internal class DefendantEntityTest {
       title = "Mr",
       familyName = "familyName",
       givenName = "givenName",
+      middleNames = listOf("middleName1", "middleName2"),
       defendantId = "defendantId",
```
markreesmoj, line 21: Not sure we really need this test - the next test src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/DefendantRepositoryIntTest.kt seems to provide coverage of this entity.

markreesmoj (approved):
LGTM

PR #185, opened 2024-02-23
CPR-187 Added additional offender fields.

vb-moj:
Refactored packages.
Refactored service.


vb-moj:
Opened another one due to a git hub issue.

PR #194, opened 2024-02-28
Cpr 187 additional fields

vb-moj:
Refactored packages.

Refactored Offender Service.


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/offender/Aliases.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.justice.digital.hmpps.personrecord.client.model.offender
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import java.time.LocalDate
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class Aliases(
```
markreesmoj, line 7: Maybe should be named `Alias` as it only holds one alias?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/offender/HighLight.kt
```diff
@@ -0,0 +1,15 @@
+package uk.gov.justice.digital.hmpps.personrecord.client.model.offender
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonProperty
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class HighLight(
```
markreesmoj, line 7: The case is a bit confusing - Highlight might be better but I suppose this is how the JSON is

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/offender/HighLight.kt
```diff
@@ -0,0 +1,15 @@
...
+  @JsonProperty("surname")
+  val preferredName: List<String>? = emptyList(),
+) {
+  companion object
+  fun getPreferredNameString(): String? {
```
markreesmoj, line 12: I don't really like the `String` on the end of this method name

markreesmoj (dismissed):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/offender/Aliases.kt
vb-moj: Agree.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/offender/HighLight.kt
vb-moj: Probably a typo , will check and change.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/offender/HighLight.kt
vb-moj: Will check if Kotlin allows to override the getter.

markreesmoj (approved):
👍

PR #197, opened 2024-03-04
CPR-131 Prison service rest client.

vb-moj:


src/main/resources/application.yml
```diff
@@ -104,6 +110,9 @@ offender-search:
 prisoner-search:
   possible-matches: /prisoner-search/possible-matches
 
+prison-service:
+  prisoner-details: /offender
+
```
vb-moj, line 115: We call it 'Prisoner Details'.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/PrisonServiceClient.kt
```diff
@@ -0,0 +1,18 @@
...
+import uk.gov.justice.digital.hmpps.personrecord.config.PrisonServiceOAuth2Config
+
+@FeignClient(
+  name = "prison-service",
+  url = "\${prison-service.base-url}",
+  configuration = [PrisonServiceOAuth2Config::class],
+)
+interface PrisonServiceClient {
+
+  @GetMapping(value = ["\${prison-service.prisoner-details}/{prisonerNumber}"])
```
markreesmoj, line 16: I know it's a pattern we have elsewhere but I am not sure why we get the path from a property. It does not change per environment. `prison-service.prisoner-details` is always `/offender`

src/test/resources/__files/prisoner-details/GET_prisoner_details_success_any_number.json
```diff
@@ -0,0 +1,257 @@
...
+  "csraClassificationDate": "2024-02-29",
+  "category": "string",
+  "categoryCode": "string",
+  "birthPlace": "WALES",
+  "birthCountryCode": "GBR",
+  "inOutStatus": "IN, OUT, TRN",
+  "identifiers": [
+    {
+      "type": "PNC",
+      "value": "1231/XX/121",
```
markreesmoj, line 120: Please can we have a valid PNC here? It may trip us up later
Here are some valid PNCs:
https://github.com/ministryofjustice/hmpps-person-record/blob/main/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/PNCIdentifierTest.kt#L76

markreesmoj (changes requested):
Looks good, but would be great to get a valid PNC in the sample data for future proofing

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/PrisonServiceClient.kt
vb-moj: It's a recommended practice. For easy maintenance, future-proofing, transparency and consistency.

src/test/resources/__files/prisoner-details/GET_prisoner_details_success_any_number.json
```diff
@@ -117,7 +117,7 @@
   "identifiers": [
     {
       "type": "PNC",
-      "value": "1231/XX/121",
+      "value": "20120052494O",
```
markreesmoj, line 120: Pretty sure this is still invalid - https://www.criminaljusticehub.org.uk/jargon-buster/police-national-computer-id-pnc-id/ says O is not a valid check letter. I think it should be a Q

markreesmoj (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/PrisonServiceClient.kt
markreesmoj: It is certainly consistent :-)!

src/test/resources/__files/prisoner-details/GET_prisoner_details_success_any_number.json
vb-moj: Yup.

markreesmoj (approved):
👍

PR #201, opened 2024-03-08
CPR-167 Added retry executor function

vb-moj:
Updated prisoner/offender services to use the retry mechanism for remote API calls.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/OffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/OffenderService.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.personrecord.service
 
 import kotlinx.coroutines.runBlocking
 import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.stereotype.Service
 import org.springframework.web.client.HttpClientErrorException
 import org.springframework.web.client.HttpServerErrorException
@@ -21,6 +22,9 @@ class OffenderService(
   private val client: ProbationOffenderSearchClient,
   private val featureFlag: FeatureFlag,
 ) {
+  @Value("\${retry.delay}")
+  private val retryDelay: Long = 0
+
   companion object {
     private val log = LoggerFactory.getLogger(this::class.java)
     const val NO_RECORDS_MESSAGE = "No Delius matching records exist"
@@ -52,7 +56,7 @@ class OffenderService(
 
   private fun getOffenderMatcher(person: Person): OffenderMatcher = runBlocking {
     try {
-      return@runBlocking RetryExecutor.runWithRetry(exceptionsToRetryOn, MAX_RETRY_ATTEMPTS) {
+      return@runBlocking RetryExecutor.runWithRetry(exceptionsToRetryOn, MAX_RETRY_ATTEMPTS, retryDelay) {
         val offenderDetails = client.findPossibleMatches(OffenderMatchCriteria.from(person))
         OffenderMatcher(offenderDetails, person)
       }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/OffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/OffenderService.kt
@@ -29,7 +29,6 @@ class OffenderService(
     const val MULTIPLE_MATCHES_MESSAGE = "Multiple Delius matches found"
     val exceptionsToRetryOn = listOf(HttpClientErrorException::class, HttpServerErrorException::class)
     const val MAX_RETRY_ATTEMPTS = 3
-
   }
 
   fun processAssociatedOffenders(personEntity: PersonEntity, person: Person) {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/OffenderService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/OffenderService.kt
@@ -1,7 +1,10 @@
 package uk.gov.justice.digital.hmpps.personrecord.service
 
+import kotlinx.coroutines.runBlocking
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PrisonerService.kt
```diff
@@ -78,24 +81,43 @@ class PrisonerService(
...
+        val pncNumbersReturned = prisoners?.joinToString(" ") { it.pncNumber.toString() }
+        log.debug("Number of prisoners returned from Nomis for PNC ${person.otherIdentifiers?.pncIdentifier} = ${prisoners?.size ?: 0} having PNCs: $pncNumbersReturned")
+        return@runBlocking PrisonerMatcher(prisoners, person)
+      } catch (exception: Exception) {
+        lastException = exception
+      }
+      delay(2000)
+    }
+    telemetryService.trackEvent(TelemetryEventType.NOMIS_CALL_FAILED, mapOf("Prison Number" to person.otherIdentifiers?.prisonNumber))
+    throw lastException ?: RuntimeException("Unexpected Error")
```
vb-moj, line 120: This is using the Kotlin CoroutineContext.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PrisonerService.kt
```diff
@@ -78,24 +81,43 @@ class PrisonerService(
...
+      throw exception
     }
+
     return prisonerDetails
   }
   private fun updatePrisonerAddresses(prisonerNumber: String, prisonerDetails: PrisonerDetails) {
-    val prisonerAddress = prisonServiceClient.getPrisonerAddresses(prisonerNumber)
+    val retryExecutor = RetryExecutor<List<Address>>()
+    val prisonerAddress = retryExecutor.executeWithRetry { prisonServiceClient.getPrisonerAddresses(prisonerNumber)!! }
     if (prisonerAddress?.isNotEmpty() == true) {
```
vb-moj, line 101: This one and the above is using a separate function based on another framework.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/RetryExecutor.kt
```diff
@@ -0,0 +1,24 @@
...
+    .build()
+
+  fun executeWithRetry(supplier: Supplier<T>): T? {
+    return try {
+      Retry.decorateSupplier(Retry.of("id", retryConfig), supplier).get()
+    } catch (e: Exception) {
+      throw RuntimeException("Failed after $MAX_ATTEMPTS retry attempts", e)
+    }
+  }
+}
```
vb-moj, line 24: [https://resilience4j.readme.io/docs/getting-started](url)

vb-moj: This is my choice!

markreesmoj:
https://github.com/ministryofjustice/hmpps-tier/blob/def934e627b60acbfe354f6049f64f6336fb3fe0/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppstier/config/RetryInterceptor.kt#L16

Not sure it is worth bringing in an external library just to do this. The above offers configurable retries and exceptions to retry on, might be worth further investigation?

markreesmoj:
Also a test to prove that both these approaches work as expected would be great

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PrisonerService.kt
markreesmoj: I don't really get coroutines :-)

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/OffenderService.kt
```diff
@@ -24,6 +27,8 @@ class OffenderService(
     const val EXACT_MATCH_MESSAGE = "Exact delius match found - adding offender to person record"
     const val PARTIAL_MATCH_MESSAGE = "Partial Delius match found"
     const val MULTIPLE_MATCHES_MESSAGE = "Multiple Delius matches found"
+    val exceptionsToRetryOn = listOf(HttpClientErrorException::class, HttpServerErrorException::class)
```
vb-moj, line 30: No list was given(not part of the story but better this way) so added the usual ones as suggested.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/RetryExecutor.kt
```diff
@@ -0,0 +1,29 @@
+package uk.gov.justice.digital.hmpps.personrecord.service
+
+import kotlinx.coroutines.delay
+import kotlin.reflect.KClass
+
+private const val DELAY: Long = 2000
+
```
markreesmoj, line 7:
```  Test should retry three times and throw the exception and send correct telemetry event() PASSED (24.1s)
  Test should track partial match event when partial match found() PASSED
  Test should retry three times before returning prisoner matcher() PASSED (4s)
  Test should not retry when exception thrown is not in the list of exceptions for retry() PASSED
  Test should add prisoner to person record when exact match found associated prisoners correctly() PASSED
  Test should retry before returning prisoner details() PASSED (6s)```
  

The tests are quite slow now, most likely  as a result of this 2s delay. 
It's not configurable at the moment, but it might be an idea to make it configurable and pass in a much smaller value for testing.
Also, the maxAttempts and exceptions are configurable but the same values are always passed in - maybe we could keep them configurable but use defaults and named arguments?

vb-moj: Thought so but it is not part of the story (even the exceptions list is not part of the story).  I am ok to make the changes but I will be on holiday from tomorrow so not sure I can finish that.

markreesmoj (approved):
👍

PR #207, opened 2024-03-15
CPR-199

markreesmoj:
Handle records from HMCTS and nDelius where the PNC corresponds to an existing record or records but no other fields being compared match

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonRecordService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonRecordService.kt
@@ -54,15 +54,19 @@ class PersonRecordService(
     offenderAliases.forEach { offenderAliasEntity -> offenderAliasEntity.offender = offenderEntity }
     offenderEntity.aliases.addAll(offenderAliases)
 
-    offenderEntity.person = personEntity
-    personEntity.offenders.add(offenderEntity)
-    return personRepository.saveAndFlush(personEntity)
+    return addOffenderToPerson(personEntity, offenderEntity)
   }
 
   fun addPrisonerToPerson(personEntity: PersonEntity, prisonerDetails: PrisonerDetails): PersonEntity {
     val prisonerEntity = PrisonerEntity.from(prisonerDetails)
     prisonerEntity.person = personEntity
     personEntity.prisoners.add(prisonerEntity)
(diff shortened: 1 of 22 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/matcher/Matcher.kt
```diff
@@ -6,7 +6,7 @@ abstract class Matcher<T>(val items: List<T>?, protected val person: Person) {
   abstract fun isMatchingItem(item: T): Boolean
   abstract fun isPartialMatchItem(item: T): Boolean
 
-  fun isExactMatch() = !items.isNullOrEmpty() && items.size == 1 && items.any(::isMatchingItem)
+  open fun isExactMatch() = !items.isNullOrEmpty() && items.size == 1 && items.any(::isMatchingItem)
```
markreesmoj, line 9: premature generalisation? We do not take updates from HMCTS so we want to create a record where we do not already have one and ignore everything else.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ProbationCaseEngagementService.kt
```diff
@@ -26,6 +26,7 @@ class ProbationCaseEngagementService(
   }
 
   private fun handlePncPresent(newOffenderDetail: DeliusOffenderDetail) {
+    // somehow find a way to test this - an update should cause a problem here if there is more than one record per PNC
```
markreesmoj, line 29: same bug will occur here I think.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonRecordService.kt
```diff
@@ -56,13 +55,6 @@ class PersonRecordService(
     return personRepository.saveAndFlush(personEntity)
   }
 
-  fun addPrisonerToPerson(personEntity: PersonEntity, prisoner: Prisoner): PersonEntity {
-    val prisonerEntity = PrisonerEntity.from(prisoner)
-    prisonerEntity.person = personEntity
-    personEntity.prisoners.add(prisonerEntity)
-    return personRepository.saveAndFlush(personEntity)
-  }
```
markreesmoj, line 64: unused

grnewbould (approved):

PR #211, opened 2024-03-20
CPR-248 Move env vars into helm for test envs

H-Iwanejko:
* Tidy up, env vars into the same file within helm rather than separate application config files


markreesmoj (approved):
:broom

PR #228, opened 2024-04-08
CPR-57

vb-moj:
Made the detailUrl optional for domain events.

Updated IT.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/processors/OffenderCreatedEventProcessor.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/processors/OffenderCreatedEventProcessor.kt
@@ -39,8 +39,11 @@ class OffenderCreatedEventProcessor(
       },
     )
   }
-  private fun getNewOffenderDetail(offenderDetailsUrl: String): Result<DeliusOffenderDetail?> = try {
-    Result.success(offenderDetailRestClient.getNewOffenderDetail(URI.create(offenderDetailsUrl).path))
+  private fun getNewOffenderDetail(offenderDetailsUrl: String?): Result<DeliusOffenderDetail?> = try {
+    Result.success(
+      offenderDetailsUrl?.let { URI.create(it).path }
+        ?.let { offenderDetailRestClient.getNewOffenderDetail(it) },
+    )
   } catch (e: Exception) {
     Result.failure(e)
   }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/DomainEvent.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/DomainEvent.kt
@@ -1,7 +1,10 @@
 package uk.gov.justice.digital.hmpps.personrecord.model
 
+import uk.gov.justice.digital.hmpps.personrecord.model.hmcts.AdditionalInformation
+
 data class DomainEvent(
   val eventType: String,
-  val detailUrl: String,
+  val detailUrl: String?,
   val personReference: PersonReference? = null,
+  val additionalInformation: AdditionalInformation?,
 )
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/integration/IntegrationTestBase.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/integration/IntegrationTestBase.kt
@@ -184,7 +184,7 @@ abstract class IntegrationTestBase {
   fun publishDeliusNewOffenderEvent(eventType: String, crn: String) {
     val crnType = PersonIdentifier("CRN", crn)
     val personReference = PersonReference(listOf(crnType))
-    val domainEvent = objectMapper.writeValueAsString(DomainEvent(eventType = eventType, detailUrl = createDetailUrl(crn), personReference = personReference))
+    val domainEvent = objectMapper.writeValueAsString(DomainEvent(eventType = eventType, detailUrl = createDetailUrl(crn), personReference = personReference, additionalInformation = null))
     val publishRequest = PublishRequest.builder().topicArn(domainEventsTopic?.arn)
       .message(domainEvent)
       .messageAttributes(


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/DomainEvent.kt
```diff
@@ -1,7 +1,10 @@
 package uk.gov.justice.digital.hmpps.personrecord.model
 
+import uk.gov.justice.digital.hmpps.personrecord.model.hmcts.AdditionalInformation
+
 data class DomainEvent(
   val eventType: String,
-  val detailUrl: String,
+  val detailUrl: String?,
```
vb-moj, line 7: Not all the domain events have this.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/processors/OffenderCreatedEventProcessor.kt
```diff
@@ -39,8 +39,11 @@ class OffenderCreatedEventProcessor(
       },
     )
   }
-  private fun getNewOffenderDetail(offenderDetailsUrl: String): Result<DeliusOffenderDetail?> = try {
-    Result.success(offenderDetailRestClient.getNewOffenderDetail(URI.create(offenderDetailsUrl).path))
+  private fun getNewOffenderDetail(offenderDetailsUrl: String?): Result<DeliusOffenderDetail?> = try {
+    Result.success(
+      offenderDetailsUrl?.let { URI.create(it).path }
```
markreesmoj, line 44: Is this now nullable?

It would be great to see a test for this new behaviour

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/processors/PrisonerUpdatedEventProcessor.kt
```diff
@@ -0,0 +1,18 @@
...
+@Component(value = PRISONER_UPDATED)
+class PrisonerUpdatedEventProcessor() : EventProcessor() {
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+  override fun processEvent(domainEvent: DomainEvent) {
+    val nomsNumber = domainEvent.additionalInformation?.nomsNumber!!
+    log.debug("Entered processEvent with  $nomsNumber")
+  }
+}
```
markreesmoj, line 18: Given what we discussed yesterday, I don't think we need separate processors for these two events as we are going to treat them in the same way (apart from logging)

src/test/resources/application-test.yml
```diff
@@ -48,7 +48,7 @@ hmpps.sqs:
     cprdeliusoffendereventsqueue:
       queueName: cpr_delius_offender_events_queue
       subscribeTopicId: domainevents
-      subscribeFilter: '{"eventType":[ "probation-case.engagement.created", "probation-case.merge.completed", "probation-case.unmerge.completed"] }'
+      subscribeFilter: '{"eventType":[ "probation-case.engagement.created", "prisoner-offender-search.prisoner.created", "prisoner-offender-search.prisoner.updated"] }'
```
markreesmoj, line 51: We should also make this change in `application-local`

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/processors/PrisonerUpdatedEventProcessor.kt
markreesmoj:
```
 if(domainEvent.eventType.equals("probation-case.engagement.created") {
      return context.getBean(domainEvent.eventType)
    }
    return PrisonerCreatedEventProcessor() // or whatever we call it.
```

This should do the trick without changing the design too much - we are just introducing complexity we will not need (unless we change course again :-)) by having 2 processors

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/processors/PrisonerUpdatedEventProcessor.kt
```diff
@@ -0,0 +1,18 @@
+package uk.gov.justice.digital.hmpps.personrecord.message.listeners.processors
+
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.personrecord.model.DomainEvent
+
+const val PRISONER_UPDATED = "probation-case.engagement.created"
```
markreesmoj, line 7: if we do end up sticking with 2 processors, this value needs changing

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/processors/PrisonerUpdatedEventProcessor.kt
vb-moj: Those two are the beans for each type. Isn't? we can have a chat.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/processors/PrisonerUpdatedEventProcessor.kt
vb-moj: Yes, copy/past error.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PrisonerService.kt
```diff
@@ -53,6 +53,10 @@ class PrisonerService(
     }
   }
 
+  fun processPrisonerDomainEvent(){
```
vb-moj, line 56: This is to fix the detect empty constructor issue.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/processors/PrisonerUpdatedEventProcessor.kt
vb-moj: In my opinion, each event type should be a bean which can still share the same service/ implementation etc..

PR #235, opened 2024-04-11
CPR-188 call splink

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/MatchScoreClient.kt
```diff
@@ -0,0 +1,40 @@
...
+  val uniqueId: MatchRequestData,
+  @JsonProperty("first_name")
+  val firstName: MatchRequestData,
+  val surname: MatchRequestData,
+  @JsonProperty("dob")
+  val dateOfBirth: MatchRequestData,
+  @JsonProperty("pnc_number")
+  val pncNumber: MatchRequestData,
+  @JsonProperty("source_dataset")
+  val sourceDataSet: MatchRequestData = MatchRequestData("libra", "delius"),
```
H-Iwanejko, line 37: Do we not need to specify which source it actually comes from here?

markreesmoj: https://github.com/ministryofjustice/hmpps-person-match-score/blob/main/hmpps_person_match_score/views/match_view.py#L52

They will be defaulted if we don't, and ignored if that comment is to be believed.

As these values are actually incorrect, it's probably best not to supply them. The next move here will be to get a custom model to use for comparisons, at which point we will write a new endpoint and probably won't use these fields at all.

H-Iwanejko (dismissed):
LGTM ✅

vb-moj (approved):

PR #236, opened 2024-04-11
CPR-165 CRO canonical + validator

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifier.kt
```diff
@@ -0,0 +1,93 @@
...
+  val fingerprint: Boolean = inputFingerprint
+
+  val valid: Boolean
+    get() = croId.isNotEmpty()
+
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (javaClass != other?.javaClass) return false
+    other as CROIdentifier
+    return croId == other.croId
```
markreesmoj, line 15: Although it is unlikely to happen in practice, this check opens up the possibility of two CROIdentifiers with the same corId and a different fingerprint being equal. Equals checks must compare all relevant fields.

```  override fun equals(other: Any?): Boolean {
    return EqualsBuilder.reflectionEquals(this, other)
  }```
  
  
  This is a failsafe way to implement equals - there is also a HashcodeBuilder which works in the same way
  
  ```https://mvnrepository.com/artifact/org.apache.commons/commons-lang3/3.14.0``` is the dependency

File: src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifier.kt [hunk 2]
```diff
@@ -0,0 +1,93 @@
...
+      val checkChar = inputCroId.takeLast(1)
+      val (yearDigits, serialNum) = inputCroId.drop(2).split(SLASH) // splits into [YY, NNNNNND]
+      val paddedSerialNum = padSerialNumber(serialNum.dropLast(1))
+      return "$paddedSerialNum/$yearDigits$checkChar"
+    }
+
+    private fun correctModulus(inputCroId: String, fingerprint: Boolean): Boolean {
+      val checkChar = inputCroId[LAST_CHARACTER]
+      val (serialNum, yearDigit) = inputCroId.dropLast(1).split(SLASH) // Append year digit to serial number
+      val serialNumToCheck = if (!fingerprint) serialNum.trimStart { it == '0' } else serialNum // Assume if no fingerprint it is SF
```
markreesmoj, line 73: these 2 comments aren't clear to me. The naming seems clear enough

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifier.kt
```diff
@@ -0,0 +1,93 @@
...
+      val modulus = VALID_LETTERS[(yearDigit + serialNumToCheck).toLong().mod(VALID_LETTERS.length)]
+      return modulus == checkChar
+    }
+
+    private fun padSerialNumber(serialNumber: String): String {
+      if (serialNumber.length < SERIAL_NUM_LENGTH) {
+        return serialNumber.padStart(SERIAL_NUM_LENGTH, '0')
+      }
+      return serialNumber
```
markreesmoj, line 82: ```return ("00000" + serialNumber).takeLast(SERIAL_NUM_LENGTH)``` 

This avoids a conditional - I got the idea from the way you use string functions in the rest of the code

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -32,6 +27,11 @@ class CourtCaseEventsService(
       is InvalidPNCIdentifier -> trackEvent(INVALID_PNC, mapOf("PNC" to pncIdentifier.invalidValue()))
       else -> trackEvent(MISSING_PNC, emptyMap())
     }
+    if (person.otherIdentifiers?.croIdentifier != null) {
+      if (!person.otherIdentifiers.croIdentifier.valid) {
+        trackEvent(INVALID_CRO, mapOf("CRO" to person.otherIdentifiers.croIdentifier.croId))
```
markreesmoj, line 32: I think this will just log an empty string if it's invalid, if I have understood `CroIdentifier` correctly. Let's put a test in CourtCaseEventsServiceIntTest to verify this.

More generally though:
I am not sure why we care about invalid CROs really - seems a lot of complexity for us to log this event. What are we proposing to do with this information?
Same goes for PNCs now - I suggest we could ditch a lot of the complexity there now we have the validation working.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifierTest.kt
```diff
@@ -0,0 +1,64 @@
...
+    assertEquals("065656/91H", identifier.croId)
+    assertEquals(true, identifier.fingerprint)
+  }
+
+  @Test
+  fun `should process CROs`() {
+    val readAllLines = Files.readAllLines(Paths.get("src/test/resources/valid_cros.csv"), Charsets.UTF_8)
+
+    readAllLines.stream().forEach {
+      println(it)
```
markreesmoj, line 60: No need to print the results out here

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/helper/HmctsMessages.kt
```diff
@@ -347,7 +347,7 @@ fun commonPlatformHearingWithNewDefendant() = """
               { 
                 "id": "b5cfae34-9256-43ad-87fb-ac3def34e2ac",
                 "pncId": "2003/0062845E",
-                "croNumber": "12345ABCDEF",
+                "croNumber": "51072/62R",
```
markreesmoj, line 350: 👍

markreesmoj (changes requested):
The equals/hashcode needs to be sorted, and either add a test for AppInsights for the invalid CROs or question the requirement for that functionality.

markreesmoj:
```> Task :detekt FAILED
/home/circleci/project/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt:12:1: uk.gov.justice.digital.hmpps.personrecord.service.type.TelemetryEventType.* is a wildcard import. Replace it with fully qualified imports. [WildcardImport]```

You can configure IntelliJ not to do * imports 

<img width="1002" alt="Screenshot 2024-04-12 at 09 03 43" src="https://github.com/ministryofjustice/hmpps-person-record/assets/76954782/0d973543-aba2-40c3-ad52-94effff2cb9c">

File: src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifier.kt (hunk 3 above)
Comment (H-Iwanejko) [reply src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifier.kt:82]: Removed the conditional as thinking about it, it is redundant as it wont pad if it is the right length anyway. Updated 👍

File: src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifier.kt (hunk 2 above)
Comment (H-Iwanejko) [reply src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifier.kt:73]: Removed 👍

File: src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifier.kt (hunk 1 above)
Comment (H-Iwanejko) [reply src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifier.kt:15]: Added 👍

File: src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifierTest.kt (hunk 5 above)
Comment (H-Iwanejko) [reply src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifierTest.kt:60]: Removed 👍

File: src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifier.kt [hunk 7]
```diff
@@ -0,0 +1,91 @@
...
+    get() = croId.isNotEmpty()
+
+  val invalidCro: String = invalidInputCro
+
+  override fun equals(other: Any?): Boolean {
+    return EqualsBuilder.reflectionEquals(this, other)
+  }
+
+  override fun hashCode(): Int {
+    return croId.hashCode()
```
markreesmoj, line 20: This needs to change to match the equals implementation. First answer on this link is good:

https://stackoverflow.com/questions/2265503/why-do-i-need-to-override-the-equals-and-hashcode-methods-in-java

```HashcodeBuilder.reflectionHashcode(this)```

File: src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifier.kt [hunk 8]
```diff
@@ -0,0 +1,91 @@
...
+    private const val VALID_LETTERS = "ZABCDEFGHJKLMNPQRTUVWXY"
+
+    private val SF_CRO_REGEX = Regex("^SF\\d{2}/\\d{1,6}[A-Z]\$")
+    private val CRO_REGEX = Regex("^\\d{1,6}/\\d{2}[A-Z]\$")
+
+    fun from(inputCroId: String? = EMPTY_CRO): CROIdentifier {
+      if (inputCroId.isNullOrEmpty()) {
+        return CROIdentifier(EMPTY_CRO, false)
+      }
+      return toCanonicalForm(inputCroId)
```
markreesmoj, line 41: minor point but I quite like the ```return when``` syntax you have used elsewhere in this class, might suit here too.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifier.kt
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private fun isExpectedFormat(inputCroId: String): Boolean {
+      return inputCroId.matches(CRO_REGEX)
+    }
+
+    private fun isSfFormat(inputCroId: String): Boolean {
+      return inputCroId.matches(SF_CRO_REGEX)
+    }
+  }
```
markreesmoj, line 90: Taken as a whole, this is considerably better than the PNCIdentifier

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/CROIdentifier.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    private fun canonicalizeSfFormat(inputCroId: String): String {
+      val checkChar = inputCroId.takeLast(1)
+      val (yearDigits, serialNum) = inputCroId.drop(2).split(SLASH) // splits into [YY, NNNNNND]
+      val paddedSerialNum = padSerialNumber(serialNum.dropLast(1))
+      return "$paddedSerialNum/$yearDigits$checkChar"
+    }
+
+    private fun correctModulus(inputCroId: String, fingerprint: Boolean): Boolean {
+      val checkChar = inputCroId[LAST_CHARACTER]
```
markreesmoj, line 72: we seem to do this three times and are often retrieving the different parts of the number. I've not thought this through, but is there anything to be said for having an interim object for validation? Something like this
```
Cro(val checkChar: String, val serialNum: String, val yearDigits: String, val fingerprint: Boolean)
```

markreesmoj (changes requested):
Hashcode needs sorting, other things are just suggestions

markreesmoj (approved):

PR #240, opened 2024-04-15
CPR-165 Add fingerprint column to tables

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PrisonerEntity.kt
```diff
@@ -114,10 +117,12 @@ class PrisonerEntity(
       )
     }
     fun from(prisonerDetails: PrisonerDetails): PrisonerEntity {
+      val croIdentifier = CROIdentifier.from(prisonerDetails.getCro())
       return PrisonerEntity(
         prisonNumber = prisonerDetails.offenderNo,
         pncNumber = PNCIdentifier.from(prisonerDetails.getPnc()),
-        cro = CROIdentifier.from(prisonerDetails.getCro()),
+        cro = croIdentifier,
+        fingerprint = croIdentifier.fingerprint,
```
vb-moj, line 125: How do we receive the fingerprint?

src/main/resources/db/migration/V14__add_fingerpint_to_tables.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
+
+ALTER table offender
+    Add column fingerprint BOOLEAN NULL;
+
+ALTER table defendant
+    Add column fingerprint BOOLEAN NULL;
+
+ALTER table prisoner
+    Add column fingerprint BOOLEAN NULL;
```
markreesmoj, line 10: I think the default value should be false

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PrisonerEntity.kt
```diff
@@ -41,6 +41,9 @@ class PrisonerEntity(
   @Convert(converter = CROIdentifierConverter::class)
   val cro: CROIdentifier? = null,
 
+  @Column(name = "fingerprint")
+  val fingerprint: Boolean? = null,
```
markreesmoj, line 45: A minor point but we do not need to name this if the column is called fingerprint. I appreciate that we have done this for all the other columns though

src/main/resources/db/migration/V14__add_fingerpint_to_tables.sql
H-Iwanejko: Done 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PrisonerEntity.kt
markreesmoj: Also can we get away with not making this nullable and defaulting to false? It will be easier to deal with

markreesmoj: Beat me to it

markreesmoj (approved):

PR #242, opened 2024-04-15
CPR-188 retry calls to match service

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/MatchScoreClient.kt
```diff
@@ -33,6 +33,8 @@ class MatchRequest(
   val dateOfBirth: MatchRequestData,
   @JsonProperty("pnc_number")
   val pncNumber: MatchRequestData,
+  @JsonProperty("source_dataset")
+  val sourceDataset: MatchRequestData = MatchRequestData("libra", "delius"),
```
markreesmoj, line 37: This is required, despite what the comment in `hmpps-person-match-score` says!

vb-moj, line 37: Are these values just for testing?

vb-moj (commented):
One minor comment.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/MatchScoreClient.kt
markreesmoj: No, they are required values. They are irrelevant to a certain extent as the same model is always used to score the match, but we have to supply them. 
This is really not much more than a spike to prove we can score records using the API using the existing endpoint. All we do is record the score in an event - it is not currently saved or used for decision making.
Next steps are to write endpoints for matching records from specific sources

markreesmoj: https://github.com/ministryofjustice/hmpps-person-match-score/blob/main/hmpps_person_match_score/views/match_view.py

vb-moj (approved):

PR #251, opened 2024-04-18
CPR-282 PNC identifier refactor

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/PNCIdentifier.kt
```diff
@@ -1,65 +1,76 @@
...
 
+import org.apache.commons.lang3.builder.EqualsBuilder
+import org.apache.commons.lang3.builder.HashCodeBuilder
 import java.time.LocalDate
 
-interface PNCIdentifier {
+class PNCIdentifier(val pncId: String) {
+
+  val valid: Boolean
+    get() = pncId.isNotEmpty()
```
markreesmoj, line 10: do we even need this method? It doesn't look as if anything is calling it apart from the test

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/PNCIdentifier.kt
```diff
@@ -1,65 +1,76 @@
...
     private const val LONG_PNC_ID_LENGTH = 10
     private const val CENTURY = 100
-    private const val LAST_CHARACTER = 12
     private const val YEAR_END = 4
 
-    private fun validOrInvalid(inputPncId: String): PNCIdentifier {
-      val pncId = toCanonicalForm(inputPncId)
+    internal const val SERIAL_NUM_LENGTH = 7
+
+    fun from(inputPncId: String? = ""): PNCIdentifier {
```
markreesmoj, line 36: could replace "" with `EMPTY_PNC` constant

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/PNCIdentifier.kt
```diff
@@ -1,65 +1,76 @@
...
   companion object {
-    fun from(inputPncId: String? = ""): PNCIdentifier {
-      if (inputPncId.isNullOrEmpty()) {
-        return MissingPNCIdentifier()
-      }
-      return validOrInvalid(inputPncId)
-    }
 
-    private const val VALID_LETTERS = "ZABCDEFGHJKLMNPQRTUVWXY"
+    private val PNC_REGEX = Regex("\\d{2,4}(/?)\\d{1,7}[A-Z]\$")
```
markreesmoj, line 26: the 7 in this is SERIAL_NUM_LENGTH

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/PNCIdentifier.kt
H-Iwanejko: Removed 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/PNCIdentifier.kt
H-Iwanejko: Done 👍

H-Iwanejko:
I'm not sure if this changed highlighted this, but if the PNC is empty on the person record and the search record it still counts it as a 'partial match' so i've added a check to only do that if its present.

This is subject to check very soon anyway.

markreesmoj (approved):
🧹

PR #252, opened 2024-04-18
CPR-286 do not call hmpps-person-match-score with null dates

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/matcher/DefendantMatcher.kt
```diff
@@ -11,7 +11,7 @@ class DefendantMatcher(defendants: List<DefendantEntity>?, person: Person) :
     return person.otherIdentifiers?.pncIdentifier == PNCIdentifier.from(item.pncNumber?.pncId) ||
       item.firstName.equals(person.givenName, true) ||
       item.surname.equals(person.familyName, true) ||
-      item.dateOfBirth == person.dateOfBirth
+      (item.dateOfBirth != null && item.dateOfBirth == person.dateOfBirth)
```
markreesmoj, line 14: This stops null dates from matching in the first place if it is the only match. Definitely a bodge - we need to do a lot better in here. 
Far too many matches

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -81,6 +82,21 @@ class CourtCaseEventsService(
...
+    if (nullDate(candidate, person)) {
+      return MatchResult("not scored, no date of birth", candidate.person?.personId.toString(), "defendantId", candidate.defendantId ?: "defendant1", "defendantId", person.defendantId ?: "defendant2")
+    }
+    return matchService.score(candidate, person)
+  }
+
+  private fun nullDate(
+    candidate: DefendantEntity,
+    person: Person,
+  ) = candidate.dateOfBirth == null || person.dateOfBirth == null
```
markreesmoj, line 98: Stops the null dates from being passed to hmpps-person-match-score

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/matcher/DefendantMatcher.kt
vb-moj: I think this will be gone next week. We don't need any of the matchers.

vb-moj (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/helper/HmctsMessages.kt
```diff
@@ -1,5 +1,7 @@
 package uk.gov.justice.digital.hmpps.personrecord.service.helper
 
+import org.testcontainers.shaded.org.bouncycastle.asn1.isismtt.x509.DeclarationOfMajority.dateOfBirth
+
```
markreesmoj, line 4: oops

PR #253, opened 2024-04-18
CPR-285 Fix for handling null pnc when receiving new offender details…

vb-moj:
… from nDelius.


H-Iwanejko (approved):

markreesmoj (approved):
👍 We get this error in dev so we will be able to see if it works within 10 minutes

PR #261, opened 2024-04-19
CPR-282 get rid of nullpointers

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
```diff
@@ -52,8 +52,8 @@ class CourtCaseEventsService(
     }
   }
 
-  private fun findByFirstNameSurname(firstName: String? = "firstName", surname: String? = "surname"): List<DefendantEntity> {
-    return defendantRepository.findAllByFirstNameAndSurname(firstName!!, surname!!)
+  private fun findByFirstNameSurname(firstName: String?, surname: String?): List<DefendantEntity> {
+    return defendantRepository.findAllByFirstNameAndSurname(firstName, surname!!)
```
H-Iwanejko, line 56: Surname doesn't need to be force unwrapped anymore? can remove `!!`. Otherwise that will cause a null pointer exception as well

H-Iwanejko (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/CourtCaseEventsService.kt
markreesmoj: I need a rest!

H-Iwanejko (approved):

PR #267, opened 2024-04-24
CPR-294 Added future flag to stop processing libra messages.

vb-moj:


H-Iwanejko (approved):

markreesmoj (commented):
I don't think a feature flag is what's needed here - I can't imagine we will turn it back on any time soon. And when we do, things will have changed quite a bit. 

I think it might be an idea to have a chat about how to approach this. There is a lot of code in there which is just not needed if we don't process Libra messages e.g. partial matching

vb-moj:
Decided to remove LIBRA stuff completely so closing this PR.

PR #278, opened 2024-04-26
CPR-317 Renamed person table to person_identifier.

vb-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/MatchService.kt
```diff
@@ -67,7 +67,7 @@ class MatchService(val matchScoreClient: MatchScoreClient, val telemetryService:
 
 class MatchResult(
   val probability: String,
-  val candidateRecordUUID: String,
+  // val candidateRecordUUID: String,
```
vb-moj, line 70: Commented out for now.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonIdentifierEntity.kt
```diff
@@ -1,43 +1,34 @@
...
 ) {
   companion object {
-    fun new(): PersonEntity {
-      val personEntity = PersonEntity(
+    fun new(): PersonIdentifierEntity {
+      val personIdentifierEntity = PersonIdentifierEntity(
         personId = UUID.randomUUID(),
       )
-      return personEntity
+      return personIdentifierEntity
```
markreesmoj, line 31: Minor but this can be a one liner

`fun new(): PersonIdentifierEntity = PersonIdentifierEntity(personId = UUID.randomUUID())`

markreesmoj (dismissed):
👍

markreesmoj:
Not sure how renaming the table will work when deploying to live with 2 pods. One to watch I suppose!

markreesmoj:
We could make it safe by creating the PersonIdentifier table rather than renaming it from Person. We wouldn't have any data in it that way, but I don't think we care too much about it

markreesmoj (approved):

PR #280, opened 2024-04-26
CPR-262 add an endpoint for triggering populate from nomis

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/prisoner/Prisoner.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.justice.digital.hmpps.personrecord.client.model.prisoner
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import jakarta.validation.constraints.NotBlank
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class Prisoner(
+  @NotBlank
+  val prisonerNumber: String,
```
vb-moj, line 9: This is how we should be using it as well. However, according to Steve, the agreed name is PrisonNumber. We should have a chat with them.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/populatefromnomis/PopulateFromNomis.kt
```diff
@@ -0,0 +1,67 @@
...
+      val prisonerNumbers = prisonServiceClient.getPrisonerNumbers(PageParams(0, pageSize))!!
+      val totalPages = prisonerNumbers.totalPages
+      var numbers = prisonerNumbers.numbers
+
+      for (page in 1..totalPages) {
+        numbers.forEach {
+          runWithRetry(retryables, retries, delayMillis) {
+            val prisoner = prisonerSearchClient.getPrisoner(it)
+            val person = PersonEntity(firstName = prisoner.firstName, middleNames = prisoner.middleNames, sourceSystem = NOMIS)
+            repository.save(person)
```
vb-moj, line 55: Could be SaveAndFlush()

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/populatefromnomis/PopulateFromNomis.kt
```diff
@@ -0,0 +1,68 @@
...
+      val totalPages = prisonerNumbers.totalPages
+      var numbers = prisonerNumbers.numbers
+
+      for (page in 1..totalPages) {
+        numbers.forEach {
+          runWithRetry(retryables, retries, delayMillis) {
+            val prisoner = prisonerSearchClient.getPrisoner(it)
+            val person = Person.from(prisoner)
+
+            repository.saveAndFlush(PersonEntity.from(person))
```
vb-moj, line 56: May be we should do a batch insert here instead.

markreesmoj: Interesting - not sure how I would enable this without making it application-wide which I am not sure we need to do.
Also I think that the bottleneck here will be the HTTP calls rather than the DB writes so I am unsure if it will make any difference. Definitely one to have in our back pocket if the performance is not where it needs to be

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/populatefromnomis/PopulateFromNomis.kt
```diff
@@ -0,0 +1,68 @@
...
+  suspend fun populate(): String {
+    populatePages()
+    return OK
+  }
+
+  suspend fun populatePages() {
+    // TODO understand scoping here, how many threads? This dispatcher will give up to 64
+    // https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-i-o.html
+    CoroutineScope(Dispatchers.IO).launch {
+      // if this call fails we will just restart the process, no need to retry
```
mikehalmamoj, line 45: So if it fails on the last call to prisoner search then it will just start again? I agree with 10 retries each call *should* get through eventually but...

markreesmoj: No, this is the initial call to get the total number of records. All the other calls are retried, seemed like a lot of trouble to wrap and test this one call too when we can just call it again as it won't have done anything.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/populatefromnomis/PopulateFromNomis.kt
```diff
@@ -0,0 +1,68 @@
...
+      val prisonerNumbers = prisonServiceClient.getPrisonerNumbers(PageParams(0, pageSize))!!
+      val totalPages = prisonerNumbers.totalPages
+      var numbers = prisonerNumbers.numbers
+
+      for (page in 1..totalPages) {
+        numbers.forEach {
+          runWithRetry(retryables, retries, delayMillis) {
+            val prisoner = prisonerSearchClient.getPrisoner(it)
```
mikehalmamoj, line 53: I know you only have a single thread for now, but it's still going to put a lot of load onto prisoner search. I wonder if it's worth guaging how bad that hit on prisoner search is going to be first - maybe just do the first few pages or something?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/populatefromnomis/PopulateFromNomis.kt
markreesmoj: Now this might be useful as we are getting 1000 records at a time - I am still inclined to run it first and see how it performs before making this change

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/populatefromnomis/PopulateFromNomis.kt
markreesmoj: limited to 4000 records for now

mikehalmamoj (approved):

vb-moj (approved):

PR #282, opened 2024-04-29
CPR-319 create new person table and entity

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonAddressEntity.kt
```diff
@@ -0,0 +1,35 @@
...
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
+import jakarta.persistence.Table
+import jakarta.persistence.Version
+
+@Entity
+@Table(name = "person_address")
+class PersonAddressEntity(
```
markreesmoj, line 16: minor but can we just call this AddressEntity? Is there going to be any other type of AddressEntity?

H-Iwanejko: Yeah as the 'AddressEntity' already exists just a way to separate them. Going to change the table back to just 'AddressEntity' once we switched to use this new relationship

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonAliasEntity.kt
```diff
@@ -0,0 +1,46 @@
...
+import jakarta.persistence.Id
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
+import jakarta.persistence.Table
+import jakarta.persistence.Version
+import java.time.LocalDate
+
+@Entity
+@Table(name = "person_alias")
+class PersonAliasEntity(
```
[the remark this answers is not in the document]
H-Iwanejko: As above comment

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/types/SourceSystemType.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.personrecord.model.types
+
+enum class SourceSystemType {
+  CPR,
+  NOMIS,
+  DELIUS,
+  HMCTS,
+}
```
markreesmoj, line 8: ❤️

PR #286, opened 2024-04-30
CPR-298 Remove most recent prisoner number

H-Iwanejko:
'most_recent_prisoner_number' is now redundant as all records with that field have a PNC


markreesmoj (approved):
👍

PR #293, opened 2024-05-02
CPR-262 move alias addition into from method

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
```diff
@@ -148,6 +148,10 @@ class PersonEntity(
         nationalInsuranceNumber = person.nationalInsuranceNumber,
         sourceSystem = person.sourceSystemType,
       )
+      val personAliases = PersonAliasEntity.fromList(person.personAliases)
+      personAliases.forEach { personAliasEntity -> personAliasEntity.person = personEntity }
+      personEntity.aliases.addAll(personAliases)
+      return personEntity
       return personEntity
```
vb-moj, line 155: 2 returns here.

markreesmoj: surprised that detekt or the linter did not find this. Good spot, thanks

vb-moj (approved):

PR #298, opened 2024-05-03
CPR-262 Handle nullable CRO prisoner search

H-Iwanejko:


markreesmoj (approved):
👍 although let's write a Deserializer for it this afternoon as per the PNCIdentifer

PR #309, opened 2024-05-08
CPR-328 Remove redundant entities

H-Iwanejko:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/populatefromnomis/PopulateFromNomisIntTest.kt
```diff
@@ -18,7 +17,6 @@ import java.util.concurrent.TimeUnit.SECONDS
 class PopulateFromNomisIntTest : IntegrationTestBase() {
 
   @Test
-  @Transactional // Is this dodgy?
```
H-Iwanejko, line 21: Remove this as it caused this test to become flaky, e.g. creating 8 instead of 7 records

PR #322, opened 2024-05-14
CPR-329 rename person tables

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/person/Person.kt
```diff
@@ -25,29 +25,13 @@ data class Person(
...
-  val nationalityCode: String? = null,
   val nationalInsuranceNumber: String? = null,
-  val observedEthnicityDescription: String? = null,
-  val selfDefinedEthnicityDescription: String? = null,
-  val homePhone: String? = null,
-  val mobile: String? = null,
-  val workPhone: String? = null,
-  val primaryEmail: String? = null,
-  val contacts: List<PersonContact> = emptyList(),
-  val address: List<PersonAddress> = emptyList(),
```
H-Iwanejko, line 50: I've removed redundant values here to simplify

markreesmoj (approved):
👍

markreesmoj:
Might be worth scaling down to 0 instances before starting the production deployment to avoid any problems with the pod running the old code?

H-Iwanejko:
Yeah that is a good point, ill raise a PR

PR #324, opened 2024-05-14
CPR-329 Re-organise processors directory

H-Iwanejko:
* Small PR just to move all processors into a more relevent directory location


markreesmoj (approved):
👍 although I don't like these processors very much, find the pattern confusing

PR #325, opened 2024-05-14
CPR-332 Re-try on locking exception

H-Iwanejko:
* Add a re-try mechanism: seems to be done similar with: https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/d8258fa5dfb0bca4d1f9e2f5cbb084d3db1b96ed/libs/messaging/src/main/kotlin/uk/gov/justice/digital/hmpps/listener/AwsNotificationListener.kt#L14


markreesmoj (dismissed):
👍

markreesmoj (approved):
👍

PR #355, opened 2024-05-28
CPR-315 Map initial NOMIS domain event to CPR person

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/prisoner/Events.kt
```diff
@@ -0,0 +1,6 @@
+package uk.gov.justice.digital.hmpps.personrecord.client.model.prisoner
+
+object Events {
+  const val PRISONER_CREATED = "prisoner-offender-search.prisoner.created"
+  const val PRISONER_UPDATED = "prisoner-offender-search.prisoner.updated"
+}
```
markreesmoj, line 6: These look like test only constants - if so they would be better in the `TEST` tree. I also don't think we need to use an `object` to store strings - they can just be defined in a file

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/types/SQSMessageType.kt
```diff
@@ -0,0 +1,5 @@
+package uk.gov.justice.digital.hmpps.personrecord.model.types
+
+object SQSMessageType {
+  const val NOTIFICATION = "Notification"
+}
```
markreesmoj, line 5: no need for an object

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/type/TelemetryEventType.kt
```diff
@@ -4,8 +4,6 @@ enum class TelemetryEventType(val eventName: String) {
   HMCTS_MESSAGE_RECEIVED("CprHMCTSMessageReceived"),
   HMCTS_PROCESSING_FAILURE("CprHMCTSMessageProcessingFailed"),
   MATCH_CALL_FAILED("CprMatchCallFailed"),
-  NOMIS_CREATE_MESSAGE_RECEIVED("CprNOMISCreateMessageReceived"),
-  NOMIS_UPDATE_MESSAGE_RECEIVED("CprNOMISUpdateMessageReceived"),
```
markreesmoj, line 8: 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/integration/MessagingMultiNodeTestBase.kt
```diff
@@ -68,7 +68,11 @@ abstract class MessagingMultiNodeTestBase : IntegrationTestBase() {
       val allEvents = telemetryRepository.findAllByEvent(event.eventName)
       val matchingEvents = allEvents?.filter {
         expected.entries.map { (k, v) ->
-          JSONObject(it.properties).get(k).equals(v)
+          val jsonObject = JSONObject(it.properties)
+          when {
+            (jsonObject.has(k)) -> jsonObject.get(k).equals(v)
+            else -> false
+          }
```
markreesmoj, line 75: What's this for?

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/test/messages/PrisonerSearchMessages.kt
```diff
@@ -0,0 +1,165 @@
+package uk.gov.justice.digital.hmpps.personrecord.test.messages
+
+fun prisonerSearchResponse(nomsNumber: String) = """
```
markreesmoj, line 3: 👍 although I think we will find that passing in an object with properties including nomsNumber becomes more useful as we add tests

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/integration/MessagingMultiNodeTestBase.kt
H-Iwanejko: This fixes an issue if you have 2 separate events with different maps. If you search for a property that don't exist on the event it throws a JSON exception. This way it checks all regardless of property map

markreesmoj (approved):
👍

PR #372, opened 2024-06-05
CPR-321 Get candidates for matching

H-Iwanejko:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonServiceIntTest.kt
```diff
@@ -0,0 +1,154 @@
...
+class PersonServiceIntTest : MessagingMultiNodeTestBase() {
+
+  @Autowired
+  private lateinit var personService: PersonService
+
+  @BeforeEach
+  override fun beforeEach() {
+    super.beforeEach()
+    // need to delete as obscure fields to search on???
+    personRepository.deleteAll()
```
markreesmoj, line 24: I know it's quite a lot (especially with the various identifiers), but I think we should consider better random generation of data as an alternative to this.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/specifications/PersonSpecification.kt
```diff
@@ -0,0 +1,78 @@
...
+        )
+        criteriaBuilder.le(
+          criteriaBuilder.function("levenshtein", Integer::class.java, criteriaBuilder.literal(it.toString()), dbDateAsString),
+          limit,
+        )
+      }
+    }
+  }
+
+  fun <T> combineSpecificationsWithOr(specifications: List<Specification<T>>): Specification<T>? {
```
H-Iwanejko, line 70: To handle multiple 'addresses'

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/court/libra/LibraCourtCaseListenerIntTest.kt
```diff
@@ -0,0 +1,147 @@
...
+import java.time.LocalDate
+import java.util.concurrent.TimeUnit.SECONDS
+
+class LibraCourtCaseListenerIntTest : MessagingMultiNodeTestBase() {
+
+  @BeforeEach
+  override fun beforeEach() {
+    super.beforeEach()
+    personRepository.deleteAll()
+    telemetryRepository.deleteAll()
```
H-Iwanejko, line 36: Even though we are leaning toward the 'non delete way of test' this seems like the path of less resistance / excessive development time after some trying to locate and verify the record creation on a record with no id's.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/court/libra/LibraCourtCaseListenerIntTest.kt
```diff
@@ -0,0 +1,147 @@
...
+    } matches { it == 0 }
+
+    await untilCallTo {
+      courtCaseEventsQueue?.sqsDlqClient?.countMessagesOnQueue(courtCaseEventsQueue!!.dlqUrl!!)?.get()
+    } matches { it == 0 }
+
+    checkTelemetry(
+      CPR_RECORD_CREATED,
+      mapOf("SOURCE_SYSTEM" to "HMCTS"),
+      times = 50,
```
H-Iwanejko, line 133: Would have to pull out every sqs message id and search for each one, if we didn't wipe the telemetry repo.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonServiceIntTest.kt
H-Iwanejko: I kept it for reasoning above, happy to chat about.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/hmcts/CourtCaseEventsProcessor.kt
```diff
@@ -88,6 +91,16 @@ class CourtCaseEventsProcessor(
...
 
+    val personEntities: List<PersonEntity> = personService.findCandidateRecords(person)
+    telemetryService.trackEvent(
+      TelemetryEventType.CPR_CANDIDATE_RECORD_SEARCH,
+      mapOf(
+        EventKeys.SOURCE_SYSTEM to SourceSystemType.HMCTS.name,
+        EventKeys.RECORD_COUNT to personEntities.size.toString(),
+        EventKeys.EVENT_TYPE to LIBRA_COURT_CASE.name,
+        EventKeys.MESSAGE_ID to sqsMessage.messageId,
+      ),
```
markreesmoj, line 102: I am not sure how best to do it, but we should put something in here which allows us to identify the matching algorithm we used
It will be hard to keep up to date.
Maybe a custom toString on the PersonSpecification??

H-Iwanejko: Under the hood it just constructs raw SQL from this it is very hard to determine which part of the statement did the matching part. So im not sure if this is possible with the way we have set it up??? we could log out the resulting SQL statement redacted...?

markreesmoj: You would have to override `hashcode` (and of course `equals`) to achieve this as the default would be the object reference and I don't think it should become a data class

markreesmoj: We need to be able to say "the average number of matches was 39 with algorithm 1, then we changed the levenshtein on date of birth to 1 and the average number went down to 5"

H-Iwanejko: Recording these rules in a confluence page for iteration purposes: https://dsdmoj.atlassian.net/wiki/spaces/PRD/pages/4938367021/Candidate+Searching+Algorithm

markreesmoj (approved):
👍 but please don't push to preprod until seeding is complete

PR #378, opened 2024-06-06
CPR-352 do not consume court messages when seeding

markreesmoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/CourtCaseEventsListenerDisabledIntTest.kt
```diff
@@ -0,0 +1,59 @@
...
+import software.amazon.awssdk.services.sns.model.PublishRequest
+import uk.gov.justice.digital.hmpps.personrecord.client.model.hmcts.MessageType.COMMON_PLATFORM_HEARING
+import uk.gov.justice.digital.hmpps.personrecord.integration.MessagingSingleNodeTestBase
+import uk.gov.justice.digital.hmpps.personrecord.model.identifiers.PNCIdentifier
+import uk.gov.justice.digital.hmpps.personrecord.test.messages.commonPlatformHearingWithSameDefendantIdTwice
+import uk.gov.justice.hmpps.sqs.countMessagesOnQueue
+import java.util.UUID.randomUUID
+
+@ActiveProfiles("seeding")
+@Disabled("This works when run in isolation but not as part of check.")
```
H-Iwanejko, line 20: Is there a not before all to switch application contexts for this test suite? Just a thought

markreesmoj: I don't know exactly what's going on here. I have an idea that another test is running in parallel with this one when check runs, and that is consuming the message. I will try to verify that by outputting the active profile in the telemetry event but won't be able to commit it.

This change was harder than I thought!

markreesmoj: I have confirmed that the code consuming the message does not have the profile `seeding` set.
So the code change is working as expected I think.
I am not going to run this today now but I would quite like to get it merged if possible so I can try the seeding tomorrow.

H-Iwanejko (dismissed):

PR #384, opened 2024-06-10
CPR-352 add indexes to help with deletions

markreesmoj:
https://dba.stackexchange.com/questions/37034/very-slow-delete-in-postgresql-workaround


H-Iwanejko (approved):

src/main/resources/db/migration/V28__add_indexes_for_child_tables.sql
H-Iwanejko, line 1: Needs updating to V29

H-Iwanejko (changes requested):

PR #386, opened 2024-06-10
CPR-352 do not store prisonNumber when processing probation events

markreesmoj:


src/main/resources/application.yml
```diff
@@ -108,7 +108,7 @@ populate-from-nomis:
     times: 10
 
 populate-from-probation:
-  page-size: 2000
```
markreesmoj, line 111: :-) worth a try to see if we can get it done by 10pm

H-Iwanejko (approved):

PR #395, opened 2024-06-13
CPR-321 Tweak search to ensure searching on populated fields

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/hmcts/CourtCaseEventsProcessor.kt
```diff
@@ -100,7 +100,7 @@ class CourtCaseEventsProcessor(
         EventKeys.RECORD_COUNT to pageablePersonEntities.totalElements.toString(),
         EventKeys.EVENT_TYPE to LIBRA_COURT_CASE.name,
         EventKeys.MESSAGE_ID to sqsMessage.messageId,
-        EventKeys.SEARCH_VERSION to "1",
+        EventKeys.SEARCH_VERSION to "1.1",
```
markreesmoj, line 103: why not 2?

H-Iwanejko: As it probably what v1 should have been

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonService.kt
```diff
@@ -102,16 +102,25 @@ class PersonService(
   private fun isCreateEvent(event: String?) = listOf(PRISONER_CREATED, NEW_OFFENDER_CREATED).contains(event)
 
   fun findCandidateRecords(person: Person): Page<PersonEntity> {
-    val postcodeSpecifications = person.addresses.map { PersonSpecification.levenshteinPostcode(it.postcode) }
+    val postcodeSpecifications = person.addresses.map {
+      PersonSpecification.isNotNull(it.postcode).and(PersonSpecification.levenshteinPostcode(it.postcode))
+    }
```
markreesmoj, line 107:
```suggestion
    val postcodeSpecifications = person.addresses.filterNot { it.postcode == null }.map {
      PersonSpecification.levenshteinPostcode(it.postcode)
    }
```

markreesmoj: How does this look vs using the PersonSpecification?

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/court/libra/LibraCourtCaseListenerIntTest.kt
```diff
@@ -99,7 +99,7 @@ class LibraCourtCaseListenerIntTest : MessagingMultiNodeTestBase() {
         "SOURCE_SYSTEM" to "HMCTS",
         "EVENT_TYPE" to LIBRA_COURT_CASE.name,
         "RECORD_COUNT" to "1000000",
-        "SEARCH_VERSION" to "1",
+        "SEARCH_VERSION" to "1.1",
```
markreesmoj, line 102: Maybe this should be a property? It might get annoying if we have to change it a lot as we add tests

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonService.kt
```diff
@@ -102,16 +102,25 @@ class PersonService(
...
-        .and(PersonSpecification.soundex(person.lastName, PersonSpecification.LAST_NAME)),
+      PersonSpecification.isNotNull(person.firstName).and(PersonSpecification.isNotNull(person.lastName))
+        .and(
+          PersonSpecification.soundex(person.firstName, PersonSpecification.FIRST_NAME)
+            .and(PersonSpecification.soundex(person.lastName, PersonSpecification.LAST_NAME)),
+        ),
+    )
+
+    val levenshteinDob = Specification.where(
+      PersonSpecification.isNotNull(person.dateOfBirth.toString())
```
markreesmoj, line 118: date of birth is nullable, so calling toString on it when it is null will return the string `null`. Have a look at this in a scratch file

`import java.time.LocalDate

val x:LocalDate? = null
val dateToString = x.toString()
println(x)
println(dateToString == null)`

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/hmcts/CourtCaseEventsProcessor.kt
markreesmoj: I just think 1,2,3 is simpler than using semver here, although as long as we can tie it back to the algorithm it doesn't matter as long as it changes each time

markreesmoj:
Can we do anything with that big test to show that this will reduce the number of matches?

H-Iwanejko:
> Can we do anything with that big test to show that this will reduce the number of matches?
It won't show any changes as we create the exactly the same person... we would have to generate loads of different people then try search

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonService.kt
H-Iwanejko: Moved the null check into criteria builder

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/court/libra/LibraCourtCaseListenerIntTest.kt
H-Iwanejko: Moved into a Property

markreesmoj (approved):
👍

PR #401, opened 2024-06-14
CPR-358 Create libra as a source system

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/hmcts/CourtCaseEventsProcessor.kt
```diff
@@ -19,7 +19,7 @@ import uk.gov.justice.digital.hmpps.personrecord.service.EventKeys
 import uk.gov.justice.digital.hmpps.personrecord.service.PersonService
 import uk.gov.justice.digital.hmpps.personrecord.service.TelemetryService
 import uk.gov.justice.digital.hmpps.personrecord.service.type.TelemetryEventType
-import uk.gov.justice.digital.hmpps.personrecord.service.type.TelemetryEventType.HMCTS_MESSAGE_RECEIVED
+import uk.gov.justice.digital.hmpps.personrecord.service.type.TelemetryEventType.COURT_MESSAGE_RECEIVED
```
markreesmoj, line 22: ❤️

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/hmcts/CourtCaseEventsProcessor.kt
```diff
@@ -63,12 +63,13 @@ class CourtCaseEventsProcessor(
...
       val person = Person.from(defendant)
       telemetryService.trackEvent(
-        HMCTS_MESSAGE_RECEIVED,
+        COURT_MESSAGE_RECEIVED,
         mapOf(
           EventKeys.PNC to person.otherIdentifiers?.pncIdentifier.toString(),
           EventKeys.CRO to person.otherIdentifiers?.croIdentifier.toString(),
           EventKeys.EVENT_TYPE to COMMON_PLATFORM_HEARING.name,
           EventKeys.MESSAGE_ID to sqsMessage.messageId,
+          EventKeys.SOURCE_SYSTEM to SourceSystemType.HMCTS.name,
```
markreesmoj, line 72: Not for now but I wonder if we should rename this to Common Platform in the future?
HMCTS is the Courts and Tribunals Service so covers both Common Platform and Libra (and much more)

markreesmoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/hmcts/CourtCaseEventsProcessor.kt
H-Iwanejko: Yeah i think that makes sense to do, it is getting slightly confusing.

PR #402, opened 2024-06-14
CPR-321 Actually search with source system

H-Iwanejko:
* Add missed AC: `AND originate from the same SourceSystemName as the newly received Person record`

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/specifications/PersonSpecification.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/specifications/PersonSpecification.kt
@@ -8,7 +8,7 @@ import java.time.LocalDate
 
 object PersonSpecification {
 
-  const val SEARCH_VERSION = "1.1"
+  const val SEARCH_VERSION = "1.2"
 
   const val PNC = "pnc"
   const val CRO = "cro"
@@ -17,6 +17,7 @@ object PersonSpecification {
   const val FIRST_NAME = "firstName"
   const val LAST_NAME = "lastName"
   const val DOB = "dateOfBirth"
+  const val SOURCE_SYSTEM = "sourceSystem"
 
   private const val POSTCODE = "postcode"
   private const val DATE_FORMAT = "YYYY-MM-DD"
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/specifications/PersonSpecification.kt
```diff
@@ -8,7 +8,7 @@ import java.time.LocalDate
 
 object PersonSpecification {
 
-  const val SEARCH_VERSION = "1.1"
+  const val SEARCH_VERSION = "1.2"
```
markreesmoj, line 11: Can we put in a graph in AppInsights showing average number of matches against search version?

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonServiceIntTest.kt
```diff
@@ -25,6 +25,39 @@ class PersonServiceIntTest : MessagingMultiNodeTestBase() {
...
+      ),
+    )
+    createPerson(
+      Person(
+        otherIdentifiers = OtherIdentifiers(pncIdentifier = PNCIdentifier.from("1981/0154257C")),
+        sourceSystemType = SourceSystemType.DELIUS,
+      ),
+    )
+
+    val personEntities = personService.findCandidateRecords(personToFind)
```
markreesmoj, line 54: If I understand this correctly, the search would never match any of the other records anyway because PNC is different. It might be an idea to make the records look more similar by supplying a bit more data, such that the test fails if you omit the sourceSystem from the comparison because the records match.

I may well be wrong but that is how the test appears to me

H-Iwanejko: Yeah thats why you never copy and paste a test, ill update it

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/specifications/PersonSpecification.kt
H-Iwanejko: Would this be worth it as we'd have to search further than 24h to get that data.

markreesmoj: I think it would be very useful. If memory serves, you can set the time range of an individual tile to be different to the dashboard. At the moment it won't tell us much but hopefully once we start getting the numbers down it will come in handy

markreesmoj (approved):
👍

PR #411, opened 2024-06-17
CPR-238 Add read me and scripts for term frequency

H-Iwanejko:


src/main/resources/db/scripts/README-term-frequency.md
```diff
@@ -0,0 +1,28 @@
+# Importing cluster data
```
markreesmoj, line 1:
```suggestion
# Importing term frequency data
```

markreesmoj (approved):
👍

PR #423, opened 2024-06-20
CPR-264 Use new match endpoint

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/MatchService.kt
```diff
@@ -3,67 +3,80 @@ package uk.gov.justice.digital.hmpps.personrecord.service
...
-      CPR_MATCH_SCORE_SUMMARY,
-      mapOf(
-        EventKeys.SOURCE_SYSTEM to newRecord.sourceSystemType.name,
-        EventKeys.HIGH_CONFIDENCE_COUNT to highConfidenceMatches.count().toString(),
-        EventKeys.LOW_CONFIDENCE_COUNT to (candidateRecords.size - highConfidenceMatches.count()).toString(),
-      ),
-    )
-    return highConfidenceMatches.sortedByDescending { candidate -> candidate.probability }
+    return highConfidenceMatches
   }
```
markreesmoj, line 32:
```suggestion
  fun findHighConfidenceMatches(candidateRecords: List<PersonEntity>, newRecord: Person): List<MatchResult> =
    candidateRecords.chunked(MAX_RECORDS).flatMap {
      collectHighConfidenceCandidates(it, newRecord)
    }
```

markreesmoj: This can be simplified using flatMap I think

markreesmoj (approved):
There's a lot going on in here, it looks OK but let's make sure we test it by sending a libra message to the queue which will have lots of matches in dev

PR #426, opened 2024-06-21
PI-2291 Update Probation Integration paged response model

marcus-bcl:
We are updating the probation-integration API to return Spring's PagedModel type, instead of the raw PageImpl type (see https://docs.spring.io/spring-data/commons/reference/repositories/core-extensions.html#core.web.page.paged-model).

This is technically a breaking change, however given this endpoint is only called on an ad hoc basis, it seemed simpler to just co-ordinate the API and client changes.


H-Iwanejko (approved):
LGTM ✅

PR #431, opened 2024-06-24
CPR-215 search via source source system

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/SearchService.kt
```diff
@@ -25,14 +21,15 @@ class SearchService(
...
-  fun findCandidateRecords(person: Person): List<MatchResult> {
-    val highConfidenceMatches = processPagedCandidates(person)
+  fun findCandidateRecords(person: Person, searchBySourceSystem: Boolean = true): List<MatchResult> {
+    val highConfidenceMatches = processPagedCandidates(person, searchBySourceSystem)
     return highConfidenceMatches.sortedByDescending { it.probability }
   }
 
-  private fun processPagedCandidates(person: Person): List<MatchResult> {
+  private fun processPagedCandidates(person: Person, searchBySourceSystem: Boolean): List<MatchResult> {
+    val query: Specification<PersonEntity> = if (searchBySourceSystem) findCandidatesBySourceSystem(person) else findCandidates(person)
```
markreesmoj, line 30: We always appear to be searching the same source system now. Do we need this conditional?

markreesmoj (approved):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/SearchService.kt
H-Iwanejko: This is for when we dont when assigning a UUID

PR #436, opened 2024-06-25
CPR-215 search for matching records

H-Iwanejko:
* Will start trying to match records on CREATE events across all source systems


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/test/messages/CourtCommonPlatformMessages.kt
```diff
@@ -50,7 +65,7 @@ private fun defendant(commonPlatformHearingSetup: CommonPlatformHearingSetup) =
   """{ 
                 "id": "${commonPlatformHearingSetup.defendantId}",
                 "masterDefendantId": "${commonPlatformHearingSetup.defendantId}",
-                "pncId": ${commonPlatformHearingSetup.pnc?.let { """ "${commonPlatformHearingSetup.pnc}" """.trimIndent() } ?: "null" },
+                ${commonPlatformHearingSetup.pnc?.let { """ "pncId": "${commonPlatformHearingSetup.pnc}", """.trimIndent() } ?: ""}
```
markreesmoj, line 68: this needs to be how it was before - the pnc is sometimes supplied as 
```
"pncId": null,
```

and we need to replicate this in a test.

It can also be supplied as

```
"pncId": "",
```

and of course

```
"pncId": "20003/089882A",
```

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonService.kt
```diff
@@ -60,6 +60,8 @@ class PersonService(
     if (isUpdateEvent(event)) {
       trackEvent(CPR_UPDATE_RECORD_DOES_NOT_EXIST, person)
     }
+    // Use result of search to determine assign of UUID CPR-271
+    searchForRecord(person, searchBySourceSystem = false)
```
markreesmoj, line 64: The boolean here is just passed through into the next method. It might be nicer to have two methods?
Something like
`searchAllSourceSystems`
`searchSameSourceSystem`

We can have a look at this later, easy to refactor and it will work as is.

markreesmoj (changes requested):
The pnc in the test setup needs to be like it was before, otherwise ✅

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/test/messages/CourtCommonPlatformMessages.kt
markreesmoj: Unless of course there is another case where the pncId is not present at all....? I haven't seen that in the test setup but I might have missed it.

markreesmoj (approved):
👍

PR #459, opened 2024-06-28
CPR-215 add candidate searching on create

H-Iwanejko:


markreesmoj (approved):
👍 but only as far as preprod

PR #465, opened 2024-07-01
CPR-215 Reduce joins on address table on queries

H-Iwanejko:
* Only does 1 join on the address table, rather than previously creating a join for each incoming postcode

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/specifications/PersonSpecification.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/specifications/PersonSpecification.kt
@@ -8,6 +8,7 @@ import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.AddressEntity
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.PersonEntity
 import java.time.LocalDate
 
+@Suppress("SpreadOperator")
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/specifications/PersonSpecification.kt
```diff
@@ -1,11 +1,14 @@
...
 
+import jakarta.persistence.criteria.Join
 import jakarta.persistence.criteria.JoinType.INNER
+import jakarta.persistence.criteria.Predicate
 import org.springframework.data.jpa.domain.Specification
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.AddressEntity
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.PersonEntity
 import java.time.LocalDate
 
+@Suppress("SpreadOperator")
```
H-Iwanejko, line 11: Use of spread required to pass predicates to `or` function

markreesmoj: Minor point but I don't mind this at all - I think we should allow this globally by disabling the applicable detekt rule

markreesmoj (approved):
Looks a lot better than what we have

PR #475, opened 2024-07-02
CPR-215 Log out uuid

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonService.kt
```diff
@@ -67,7 +67,7 @@ class PersonService(
       trackEvent(
         CPR_CANDIDATE_RECORD_FOUND_UUID,
         person,
-        mapOf(EventKeys.UUID to ""),
+        mapOf(EventKeys.UUID to personEntity.personIdentifier?.personId?.toString()),
```
markreesmoj, line 70: Are UUIDs being assigned yet? We will have loads in preproduction because of the data seeding, but do we have them in production?

H-Iwanejko: Not until 271 is played. This is just in preparation.

markreesmoj (approved):
👍

PR #477, opened 2024-07-03
CPR-367 Add cert to docker file and add properties

H-Iwanejko:


src/main/resources/application-postgres.properties
```diff
@@ -0,0 +1,3 @@
+spring.datasource.url=jdbc:postgresql://${DATABASE_ENDPOINT}/${DATABASE_NAME}?sslmode=verify-full
+spring.datasource.username=${DATABASE_USERNAME}
+spring.datasource.password=${DATABASE_PASSWORD}
```
markreesmoj, line 3: There is an existing `spring.datasource.url` in `application.yml` (where everything is specified in the URL).
We can split it into separate properties like this by all means - it's easier to read

markreesmoj: As it stands this file won't be loaded so these properties won't be set. We can either set a the spring profile to `postgres` in all environments, or move this into `application.yml`

If we do end up keeping it in a separate file, we should make it `yml` so all our properties are in the same format

markreesmoj (approved):
👍

src/main/resources/application-postgres.properties
H-Iwanejko: Update 👍 We can move out into seperate files in subsequent updates

PR #491, opened 2024-07-08
CPR-312 Use levenshtein less equal func

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/specifications/PersonSpecification.kt
```diff
@@ -52,7 +52,7 @@ object PersonSpecification {
         val addressJoin: Join<PersonEntity, AddressEntity> = root.join("addresses", INNER)
         val postcodePredicates: Array<Predicate> = postcodes.map {
           criteriaBuilder.le(
```
markreesmoj, line 54: looks like we have 2 l(t)e here. Do we still need the CriteriaBuilder one?

H-Iwanejko: According to the docs we still need to evaluate the result of the function: `levenshtein_less_equal returns the correct distance; otherwise it returns some value greater than max_d`

PR #495, opened 2024-07-10
CPR-347 set correct property with jwt URL

markreesmoj:


src/main/resources/application-local.yml
```diff
@@ -37,3 +37,8 @@ hmpps.sqs:
       arn: arn:aws:sns:eu-west-2:000000000000:courtcaseeventstopic
     domainevents:
       arn: arn:aws:sns:eu-west-2:000000000000:domainevents
+
+spring:
+  datasource:
+    url: jdbc:postgresql://localhost:5432/postgres?currentSchema=personrecordservice&user=root&password=dev&stringtype=unspecified
+
```
markreesmoj, line 44: This is to make the application run locally

markreesmoj: Mirrors the change made in application-test to not use SSL for the database connection when running locally

PR #511, opened 2024-07-19
CPR-395 experimenting with JPA save vs saveAndFLush

markreesmoj:
Based on reading https://thorben-janssen.com/spring-data-jpa-save-saveandflush-and-saveall/

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonService.kt
@@ -82,15 +82,8 @@ class PersonService(
   }
 
   private fun updateExistingPersonEntity(person: Person, personEntity: PersonEntity) {
-    removeAllChildEntities(personEntity)
     personEntity.update(person)
     personRepository.save(personEntity)
-  }
-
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonService.kt
```diff
@@ -82,16 +82,15 @@ class PersonService(
...
+    personEntity.update(person)
+    personRepository.save(personEntity)
   }
 
-  private fun removeAllChildEntities(personEntity: PersonEntity): PersonEntity {
+  private fun removeAllChildEntities(personEntity: PersonEntity) {
     personEntity.aliases.clear()
     personEntity.addresses.clear()
     personEntity.contacts.clear()
-    return personRepository.saveAndFlush(personEntity)
```
markreesmoj, line 94: Don't see why we have to write this change to the database at this point

H-Iwanejko:
Yeah i've had a slight read into it aswell, i've seen if you pair the `save` with the `transactional` annotation i believe it merges the operations together. Might be worth trying it/

markreesmoj:
We can consider making a transactional service at some point. The other refactor to move the clearing of the child entities into PersonEntity has already been done

PR #512, opened 2024-07-19
CPR-377 Add special characteristics to person table

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
```diff
@@ -69,8 +69,20 @@ class PersonEntity(
   @Column(name = "prison_number")
   var prisonNumber: String? = null,
 
-  @Column(name = "offender_id")
-  val offenderId: String? = null,
+  @Column(name = "birth_place")
```
markreesmoj, line 72: it's a very minor point and hardly matters, but to be consistent with the naming convention, this should be "birthplace" or we should rename the field to `birthPlace`

markreesmoj (approved):
👍

PR #514, opened 2024-07-22
CPR-373 add address fields

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/AddressEntity.kt
```diff
@@ -27,9 +27,15 @@ class AddressEntity(
...
   var person: PersonEntity? = null,
 
+  @Column(name = "address_full")
+  val fullAddress: String? = null,
+
   @Column
   val postcode: String? = null,
 
+  @Column(name = "address_type")
+  val type: String? = null,
```
H-Iwanejko, line 37: Will become an enum when the types have actually been defined

markreesmoj (approved):
👍

PR #519, opened 2024-07-23
CPR-92 Add test redrive-policy

H-Iwanejko:
Uses: https://github.com/gradle/test-retry-gradle-plugin. Should mitigate some flakeyness in some tests.


markreesmoj (approved):
👍

PR #525, opened 2024-07-29
Cpr 349 update nomis fields

JunaedAhmedMOJ:


H-Iwanejko (commented):
Could update one of the tests for this to show each field actually being populated off a nomis message

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/test/RandomTestData.kt
```diff
@@ -34,6 +34,12 @@ fun randomCro(): String {
 fun randomName(): String = randomLowerCaseString()
 fun randomNationalInsuranceNumber(): String = UUID.randomUUID().toString()
 fun randomDriverLicenseNumber(): String = UUID.randomUUID().toString()
+fun randomBirthPlace(): String = randomLowerCaseString()
+fun randomBirthCountry(): String = randomLowerCaseString()
+fun randomNationality(): String = randomLowerCaseString()
+fun randomReligon(): String = randomLowerCaseString()
+fun randomSexualOrientation(): String = randomLowerCaseString()
+fun randomSex(): String = randomLowerCaseString()
```
markreesmoj, line 42: are these all free text strings, or are some of them codes? eg is birthCountry ISO country code? It will still be a string to us of course

JunaedAhmedMOJ: have removed birth country / birth place as it was not in the response for prisoner search

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/AddressEntity.kt
```diff
@@ -44,7 +44,7 @@ class AddressEntity(
       if (address.postcode.isNullOrEmpty()) {
         return null
```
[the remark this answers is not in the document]
JunaedAhmedMOJ: will need to confirm this with BA on whether or not we will want to populate address if postcode is missing/null

PR #526, opened 2024-07-29
CPR-362 call matcher self match score

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/SearchService.kt
```diff
@@ -38,6 +38,8 @@ class SearchService(
     return matches.firstOrNull()?.candidateRecord
   }
 
+  fun retrieveRecordSelfMatchScore(person: Person): Double = matchService.getSelfMatchScore(person)
```
H-Iwanejko, line 41: It is here as well as we will need to apply a threshold filter to it in the future

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/MatchScoreClient.kt
```diff
@@ -38,4 +41,24 @@ class MatchRecord(
...
   val pnc: String? = "",
-)
+) {
+  companion object {
+    fun from(newRecord: Person): MatchRecord {
+      return MatchRecord(
+        firstName = newRecord.firstName,
+        lastname = newRecord.lastName,
+        dateOfBirth = newRecord.dateOfBirth?.toString(),
+        pnc = newRecord.references.firstOrNull { it.identifierType == IdentifierType.PNC }?.identifierValue,
```
markreesmoj, line 51: person.getPnc() would be more convenient - maybe an extension function would help

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/MessagingMultiNodeTestBase.kt
```diff
@@ -43,6 +43,8 @@ import java.util.UUID
 @ExtendWith(MultiApplicationContextExtension::class)
 abstract class MessagingMultiNodeTestBase : IntegrationTestBase() {
 
+  internal var shouldStubSelfMatchScore: Boolean = true
```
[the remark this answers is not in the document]
markreesmoj: What is it used for? It always seems to be true in this code sample

markreesmoj (approved):
Looks OK, couple of suggestions

PR #537, opened 2024-07-30
Pic 4008 additional tests

sameermoj:


markreesmoj:
We do not think that sharedTime is actually present on the messages by this point, looks like CourtHearingEventReceiver has already removed it from the message received from Common Platform

PR #538, opened 2024-07-31
Cpr 395 store messages

markreesmoj:
temporarily store message bodies so we can see why they are not being deduplicated

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/CourtMessageEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/CourtMessageEntity.kt
@@ -1,0 +1,27 @@
+package uk.gov.justice.digital.hmpps.personrecord.jpa.entity
+
+import jakarta.persistence.Column
+import jakarta.persistence.Entity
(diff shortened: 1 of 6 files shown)


markreesmoj:
I don't think we need to do this - FIFO deduplication seems to be working fine

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/CourtMessageEntity.kt
```diff
@@ -0,0 +1,27 @@
...
+@Table(name = "court_message")
+class CourtMessageEntity(
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val id: Long? = null,
+
+  @Column(name = "message_id")
+  val messageId: String? = null,
+
+  @Column(name = "message")
```
markreesmoj, line 21: Add hearingId here
Also store FIFO messages
Compare hearings in both

markreesmoj: compare to PicCourtCaseCreated

PR #541, opened 2024-07-31
CPR-376 Small refactor

H-Iwanejko:
* make sure null pnc is sent as empty string


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/MatchScoreClient.kt
```diff
@@ -57,7 +59,7 @@ data class MatchRecord(
         firstName = personEntity.firstName,
         lastname = personEntity.firstName,
         dateOfBirth = personEntity.dateOfBirth?.toString(),
-        pnc = personEntity.references.firstOrNull { it.identifierType == IdentifierType.PNC }?.identifierValue,
+        pnc = personEntity.references.getType(IdentifierType.PNC).firstOrNull()?.identifierValue ?: "",
```
markreesmoj, line 62: Can we write a test to prove that this is necessary?
Also maybe consider a convenience function like `personEntity.getPNC()`

H-Iwanejko:
Closing as null pnc doesn't matter.
Also, connivence function would be obsolete when start passing multiple pncs to matcher

PR #546, opened 2024-08-01
CPR-349-map-dl-nino

JunaedAhmedMOJ:
mapped national insurance number and drivers license to identifiers


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/prisoner/Identifier.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.personrecord.client.model.prisoner
```
grnewbould, line 1: Just wondering why these are in a prisoner package, are they not used for all people?

markreesmoj: this format is specific to prisoner-search API as far as I know

markreesmoj: https://prisoner-search-dev.prison.service.justice.gov.uk/swagger-ui/index.html?configUrl=/v3/api-docs/swagger-config#/Popular/findByPrisonNumber

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/person/Person.kt
```diff
@@ -141,6 +141,9 @@ data class Person(
       val references = listOf(
         Reference.from(IdentifierType.CRO, prisoner.cro?.toString()),
         Reference.from(IdentifierType.PNC, prisoner.pnc?.toString()),
+        Reference.from(IdentifierType.NATIONAL_INSURANCE_NUMBER, prisoner.identifiers.find { it.type == "NINO" }?.value),
```
H-Iwanejko, line 144: Could add a helper function like the person getType for identifiers for the prisoner references e.g.
```
fun List<Reference>.getType(type: IdentifierType): List<Reference> {
      return this.filter { it.identifierType == type }
    }
```

H-Iwanejko (approved):

PR #585, opened 2024-08-23
CPR-342 Merge from different UUID's

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/MergeService.kt
```diff
@@ -76,25 +104,34 @@ class MergeService(
...
       mapOf(
-        EventKeys.TO_UUID to targetPersonEntity.personKey?.let { it.personId.toString() },
-        EventKeys.FROM_UUID to sourcePersonEntity?.personKey?.let { it.personId.toString() },
-        EventKeys.TO_SOURCE_SYSTEM to targetPersonEntity.sourceSystem.name,
-        EventKeys.FROM_SOURCE_SYSTEM to sourcePersonEntity?.sourceSystem?.name,
+        EventKeys.TO_UUID to initialTargetUuid,
+        EventKeys.FROM_UUID to initialSourceUuid,
+        mergeEvent.sourceSystemId.first to mergeEvent.sourceSystemId.second,
+        mergeEvent.targetSystemId.first to mergeEvent.targetSystemId.second,
+        EventKeys.FROM_SOURCE_SYSTEM to mergeEvent.mergedRecord.sourceSystemType.name,
```
markreesmoj, line 123: Is this correct? Looks like FROM_SOURCE_SYSTEM will always be equal to  TO_SOURCE_SYSTEM, in which case why log both?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/MergeService.kt
```diff
@@ -76,25 +104,34 @@ class MergeService(
...
+        EventKeys.TO_SOURCE_SYSTEM to mergeEvent.mergedRecord.sourceSystemType.name,
       ),
     )
   }
 
+  private fun removeUuidLinkFromRecord(entity: PersonEntity) {
+    entity.personKey?.personEntities?.remove(entity)
+    entity.personKey = null
+    personRepository.saveAndFlush(entity)
```
markreesmoj, line 132: There seem to be a lot of calls to saveAndFlush in here - do we need them all?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/MergeService.kt
H-Iwanejko: Yeah fair point! Ill update in subsequent PR

PR #595, opened 2024-09-02
CPR-440 Add currently managed to person record

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/court/commonplatform/Defendant.kt
```diff
@@ -4,10 +4,10 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize
 import jakarta.validation.Valid
+import uk.gov.justice.digital.hmpps.personrecord.client.model.prisoner.deserializers.CROIdentifierDeserializer
```
markreesmoj, line 7: not sure about these being in a `prisoner` package - CROs are not only present on prisoners, same goes for PNCs

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/prisoner/CurrentlyManaged.kt
```diff
@@ -0,0 +1,26 @@
...
+      INACTIVE_TRANSFER("Inactive: TRN"),
+      INACTIVE_OUT("Inactive: OUT"),
+    }
+
+    fun from(inputStatus: String? = EMPTY_STATUS): CurrentlyManaged {
+      val status: Boolean? = when (inputStatus) {
+        Status.ACTIVE_IN.status -> true
+        Status.ACTIVE_OUT.status -> true
+        Status.INACTIVE_TRANSFER.status -> true
+        Status.INACTIVE_OUT.status -> false
```
markreesmoj, line 20: what is the difference between `false` and `null`?

markreesmoj (commented):
Looks OK, but do we really need to distinguish between false and null if someone is not currently managed? It would be simpler if it was just true/false

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/court/commonplatform/Defendant.kt
H-Iwanejko: Moved back 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/prisoner/CurrentlyManaged.kt
H-Iwanejko: I believe as its only for prisoner reocrds we dont wont the rest of the records to hold any value as per the AC's on: [this](https://dsdmoj.atlassian.net/browse/CPR-440)

markreesmoj: I am questioning the ACs themselves. What information do we gain from `null`?

Also `currently managed`is a concept in probation as well as prison (naturally enough).
Are we ever going to store that?

H-Iwanejko: Yeah there was a discussion around this. I dont recall most of it but we know for sure with a certain level for prison that they are within our care. But under probation is subjective. Probably worth asking Tony to clarify this.

markreesmoj (approved):
👍

PR #609, opened 2024-09-13
CPR-453 Use postgres similarity to reduce number records on search

H-Iwanejko:
* Use postgres https://www.postgresql.org/docs/current/pgtrgm.html trigram function to reduce the number of rows passed to the Levenshtein function. Includes indexes on the frequently used searched columns. Hopefully, speeding up the query / message processing.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
@@ -93,6 +93,9 @@ class PersonEntity(
   @Column(name = "date_of_birth")
   var dateOfBirth: LocalDate? = null,
 
+  @Column(name = "date_of_birth_text")
+  var dateOfBirthText: String? = null,
+
   @Column
   val sex: String? = null,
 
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
```diff
@@ -93,6 +93,9 @@ class PersonEntity(
   @Column(name = "date_of_birth")
   var dateOfBirth: LocalDate? = null,
 
+  @Column(name = "date_of_birth_text")
```
H-Iwanejko, line 96: Add a text column for date, as you cant have a gin index on date data type. As well as not allowing 'to_char' function when creating a index. So best option to store as text as well and run fuzzy search on that column instead

markreesmoj (approved):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
H-Iwanejko: Yeah good shout, i'll monitor the results of this to see if its worth keeping. If so, ill swap to view

PR #610, opened 2024-09-13
Revert "CPR-453 Use postgres similarity to reduce number records on search"

H-Iwanejko:
Reverts ministryofjustice/hmpps-person-record#609


markreesmoj (commented):
The migration V60 has been applied in dev (and preprod?) now, you can't just delete it

Options, neither great, are to manually delete V60 from the flyway_history table (or whatever it's called), then you can merge this, or to write another migration undoing it as V61, whichever you feel is safer

H-Iwanejko:
Yeah its not great, i was going to delete the flyway history and revert this 👍

markreesmoj (approved):
👍

PR #613, opened 2024-09-16
CPR-350-map-unmerge-delius

JunaedAhmedMOJ:
map unmerges for delius

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/probation/ProbationUnmergeEventProcessor.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/probation/ProbationUnmergeEventProcessor.kt
@@ -2,7 +2,6 @@ package uk.gov.justice.digital.hmpps.personrecord.message.processors.probation
 
 import org.springframework.stereotype.Component
 import uk.gov.justice.digital.hmpps.personrecord.client.model.sqs.domainevent.DomainEvent
-import uk.gov.justice.digital.hmpps.personrecord.jpa.repository.PersonRepository
 import uk.gov.justice.digital.hmpps.personrecord.model.types.SourceSystemType.DELIUS
 import uk.gov.justice.digital.hmpps.personrecord.service.EventKeys
 import uk.gov.justice.digital.hmpps.personrecord.service.TelemetryService
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/probation/ProbationUnmergeEventProcessor.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/probation/ProbationUnmergeEventProcessor.kt
@@ -1,0 +1,37 @@
+package uk.gov.justice.digital.hmpps.personrecord.message.processors.probation
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.personrecord.client.model.sqs.domainevent.DomainEvent
+import uk.gov.justice.digital.hmpps.personrecord.jpa.repository.PersonRepository
+import uk.gov.justice.digital.hmpps.personrecord.model.types.SourceSystemType.DELIUS
+import uk.gov.justice.digital.hmpps.personrecord.service.EventKeys
+import uk.gov.justice.digital.hmpps.personrecord.service.TelemetryService
+import uk.gov.justice.digital.hmpps.personrecord.service.type.TelemetryEventType.UNMERGE_MESSAGE_RECEIVED
+
+@Component
+class ProbationUnmergeEventProcessor(
+  val telemetryService: TelemetryService,
+  val personRepository: PersonRepository,
+) : BaseProbationEventProcessor() {
+
+  fun processEvent(domainEvent: DomainEvent) {
+    telemetryService.trackEvent(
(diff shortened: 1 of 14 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/MessagingMultiNodeTestBase.kt
```diff
@@ -175,6 +175,27 @@ abstract class MessagingMultiNodeTestBase : IntegrationTestBase() {
       ),
     )
   }
+  fun probationUnmergeEventAndResponseSetup(
```
H-Iwanejko, line 178: This is exactly the same as the function above

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationUnmergeEventListenerIntTest.kt
```diff
@@ -0,0 +1,108 @@
...
+        ),
+      ),
+    )
+
+    probationMergeEventsQueue!!.sqsClient.purgeQueue(
+      PurgeQueueRequest.builder().queueUrl(probationMergeEventsQueue!!.queueUrl).build(),
+    ).get()
+    probationMergeEventsQueue!!.sqsDlqClient!!.purgeQueue(
+      PurgeQueueRequest.builder().queueUrl(probationMergeEventsQueue!!.dlqUrl).build(),
+    ).get()
```
markreesmoj, line 93: why are we purging the queue here?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/probation/ProbationUnmergeEventProcessor.kt
```diff
@@ -0,0 +1,37 @@
...
+        EventKeys.EVENT_TYPE to domainEvent.eventType,
+        EventKeys.SOURCE_SYSTEM to DELIUS.name,
+      ),
+    )
+    getProbationCase(domainEvent.additionalInformation?.targetCrn!!).fold(
+      onSuccess = {
+        log.info("Successfully mapped unmerged record")
+      },
+      onFailure = {
+        log.error("Error retrieving offender detail: ${it.message}")
```
markreesmoj, line 32: This is a situation where we cannot find a record for the person we are trying to unmerge? We should probably make that clearer in the error message - the above is a bit generic

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/probation/ProbationUnmergeEventProcessor.kt
```diff
@@ -0,0 +1,37 @@
...
+import uk.gov.justice.digital.hmpps.personrecord.jpa.repository.PersonRepository
+import uk.gov.justice.digital.hmpps.personrecord.model.types.SourceSystemType.DELIUS
+import uk.gov.justice.digital.hmpps.personrecord.service.EventKeys
+import uk.gov.justice.digital.hmpps.personrecord.service.TelemetryService
+import uk.gov.justice.digital.hmpps.personrecord.service.type.TelemetryEventType.UNMERGE_MESSAGE_RECEIVED
+
+@Component
+class ProbationUnmergeEventProcessor(
+  val telemetryService: TelemetryService,
+  val personRepository: PersonRepository,
```
markreesmoj, line 14: This dependency isn't used

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/probation/ProbationUnmergeEventProcessor.kt
```diff
@@ -0,0 +1,37 @@
...
+) : BaseProbationEventProcessor() {
```
markreesmoj, line 15: I know this hasn't been introduced in this PR but I am not over keen on using inheritance like this. `getProbationCase` could just as easily be in a service, which would make it available for reuse should anything which isn't a ProbationEventProcessor wish to use it.

markreesmoj: There are loads of articles about favouring composition over inheritance around, and I am sure you are familiar with the pros and cons. I have generally found that it can be useful in tests (where refactoring and new development is rare) and restrictive in the `main` source tree where you need to be flexible and esponse.

JunaedAhmedMOJ: Discussed with mark, potentially need to do something around the base event classes down the line as they could cause issue in the future

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationUnmergeEventListenerIntTest.kt
JunaedAhmedMOJ: will revisit this part when doing the processing for unmerges as we will need to edit / add some new test also

markreesmoj (approved):
👍

PR #640, opened 2024-10-03
CPR-397 unmerge self match exclude

H-Iwanejko:


H-Iwanejko:
* Not adding override marker yet as will do as part of upsert in future tickets

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/MessagingMultiNodeTestBase.kt
```diff
@@ -363,6 +364,17 @@ abstract class MessagingMultiNodeTestBase : IntegrationTestBase() {
     )
   }
 
+  fun blitz(actionCount: Int, threadCount: Int, action: () -> Unit) {
```
markreesmoj, line 367: ❤️

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/OverrideMarkerEntity.kt
```diff
@@ -0,0 +1,40 @@
...
+import uk.gov.justice.digital.hmpps.personrecord.model.types.OverrideMarkerType
+
+@Entity
+@Table(name = "override_marker")
+class OverrideMarkerEntity(
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val id: Long? = null,
+
+  @ManyToOne(optional = false)
```
markreesmoj, line 23: I have no context here at all so feel free to disregard this.
At first glance I would have imagined this to be a column on either Person orPersonKey

JunaedAhmedMOJ (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/OverrideMarkerEntity.kt
H-Iwanejko: My impression is that we are only unmerging record (i think) which has a knock on effect of creating new uuids potentially. One to double check. Hence why i've only done it at record level for now

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/probation/ProbationMergeEventProcessor.kt
```diff
@@ -18,7 +19,7 @@ class ProbationMergeEventProcessor(
   val mergeService: MergeService,
 ) : BaseProbationEventProcessor() {
 
-  fun processEvent(domainEvent: DomainEvent) {
+  fun processEvent(domainEvent: DomainEvent) = runBlocking {
```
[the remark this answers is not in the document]
markreesmoj: I know that you have refactored this since, but I wanted to explain what I meant yesterday.

This is one of the downsides of using inheritance rather than composition - the move to coroutines is required in one of the subclasses but is forced on all of them. As a result, changes have to be made to classes which are not related to the work in hand and are (possibly) not getting any benefits from the change. This is a sign that there is too much coupling in the codebase.

I'm sure you have heard "favour composition over inheritance" before. I personally have found inheritance to be useful in test classes, and I think this is because they very rarely change functionality, just add or remove test methods, so refactors which affect classes which you don't need to change are unlikely. Inheritance does have its place, especially in polymorphism.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/MergeService.kt
```diff
@@ -150,17 +150,10 @@ class MergeService(
 
   private fun sourcePersonKeyHasMultipleRecords(sourcePersonEntity: PersonEntity?): Boolean = (sourcePersonEntity?.personKey?.personEntities?.size ?: 0) > 1
```
markreesmoj, line 152: I don't really understand this yet, but I really like the look of it!

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/UnmergeService.kt
```diff
@@ -0,0 +1,84 @@
...
+import uk.gov.justice.digital.hmpps.personrecord.service.type.TelemetryEventType.CPR_SELF_MATCH
+
+@Service
+class UnmergeService(
+  private val telemetryService: TelemetryService,
+  private val matchService: MatchService,
+  @Value("\${retry.delay}") private val retryDelay: Long,
+) {
+
+  private val retryExceptions = listOf(
```
markreesmoj, line 29: opportunity for code reuse here? This list seems to crop up a lot

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/OverrideMarkerEntity.kt
markreesmoj: OK, as long as you are sure that it is a ManyToOne relationship, given that it is at record (person) level. Surely a person is either included or excluded (or neither).

I am aware I am coming in like a seagull on this, sorry about that. A lot going on on PiC at the moment, just trying to help.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/probation/ProbationMergeEventProcessor.kt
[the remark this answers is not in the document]
H-Iwanejko: Yeah I agree, i think its one to sort our in a future PR as this is getting quite big now

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/UnmergeService.kt
H-Iwanejko: Move to a reusable component

JunaedAhmedMOJ (approved):

PR #641, opened 2024-10-07
CPR-428 Remove person specification

H-Iwanejko:


JunaedAhmedMOJ (approved):

markreesmoj:
https://github.com/ministryofjustice/hmpps-person-record/blob/2dd2f294f13de3d6576db52ba1503dcc6c671a80/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonRepository.kt#L11C30-L11C69

I think we no longer need PersonRepository to extend the JpaSpecificationExecutor

PR #642, opened 2024-10-07
CPR-428 Remove JPA specification executor

H-Iwanejko:


markreesmoj (approved):
🧹

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonKeyRepository.kt
```diff
@@ -1,12 +1,11 @@
 package uk.gov.justice.digital.hmpps.personrecord.jpa.repository
 
 import org.springframework.data.jpa.repository.JpaRepository
-import org.springframework.data.jpa.repository.JpaSpecificationExecutor
 import org.springframework.stereotype.Repository
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.PersonKeyEntity
 import java.util.*
```
markreesmoj, line 6: 😢 We are only importing UUID from this package, I think

PR #645, opened 2024-10-08
CPR-472-memory-fix

JunaedAhmedMOJ:
applied pagination


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/SearchService.kt
```diff
@@ -38,8 +39,14 @@ class SearchService(
 
   private fun searchForRecords(person: Person, personQuery: PersonQuery): List<MatchResult> {
     val highConfidenceMatches = mutableListOf<MatchResult>()
-    val matchCandidates = personRepository.findMatchCandidates(person, personQuery.query)
-    val totalElements = matchCandidates.size
+
+    val pageable = PageRequest.of(0, PAGE_SIZE)
```
H-Iwanejko, line 43: This will only take the first page of 50 Results. Will not then process other pages

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/SearchService.kt
```diff
@@ -38,8 +39,14 @@ class SearchService(
...
+
+    val matchCandidatesPage = personRepository.findMatchCandidates(person, personQuery.query, pageable)
+
```
H-Iwanejko, line 46: Need to loop through each batch of results

H-Iwanejko (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonBlockingRulesRepository.kt
```diff
@@ -11,10 +14,16 @@ class PersonBlockingRulesRepository {
...
+  fun findMatchCandidates(person: Person, personQuery: String, pageable: Pageable): Page<PersonEntity> {
     val query = entityManager!!.createNativeQuery(personQuery, PersonEntity::class.java)
     query.setParameter("firstName", person.firstName)
     query.setParameter("lastName", person.lastName)
-    return query.resultList as List<PersonEntity>
+
+    // apply pagination
+    query.setFirstResult((pageable.offset.toInt()))
+
+    val results = query.resultList as List<PersonEntity>
```
H-Iwanejko, line 25: This still returns a results more than the pageSize.

H-Iwanejko: I believe you need something like: 
```
query.firstResult = pageable.offset.toInt()
query.maxResults = pageable.pageSize
```

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonBlockingRulesRepository.kt
```diff
@@ -11,10 +14,16 @@ class PersonBlockingRulesRepository {
...
+
+    return PageImpl(results, pageable, results.size.toLong())
```
H-Iwanejko, line 27: The `results.size.toLong()` needs to be the total amount of results we would get without paging. Where a `COUNT (*)` statement would come into play here

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonBlockingRulesRepository.kt
```diff
@@ -11,10 +14,31 @@ class PersonBlockingRulesRepository {
...
+    val results = query.resultList as List<PersonEntity>
+
+    val countQuery = countMatchCandidates(personQuery, person)
+    return PageImpl(results, pageable, countQuery)
+  }
+
+  fun countMatchCandidates(personQuery: String, person: Person): Long {
+    val queryString = "SELECT COUNT (*) FROM ($personQuery) AS total"
+
+    println("executing count query " + queryString)
```
H-Iwanejko, line 35: Needs removing

PR #646, opened 2024-10-09
CPR-472-add-paging

JunaedAhmedMOJ:
added paging to hopefully resolve memory isue

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonBlockingRulesRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonBlockingRulesRepository.kt
@@ -15,7 +15,9 @@ class PersonBlockingRulesRepository {
   private val entityManager: EntityManager? = null
 
   fun findMatchCandidates(searchCriteria: PersonSearchCriteria, personQuery: String, pageable: Pageable): Page<PersonEntity> {
+    val countQuery = countMatchCandidates(personQuery, searchCriteria)
     val query = entityManager!!.createNativeQuery(personQuery, PersonEntity::class.java)
+
     query.setParameter("firstName", searchCriteria.firstName)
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonBlockingRulesRepository.kt
```diff
@@ -11,10 +14,29 @@ class PersonBlockingRulesRepository {
...
     query.setParameter("lastName", searchCriteria.lastName)
-    return query.resultList as List<PersonEntity>
+
+    // apply pagination
+    query.setFirstResult((pageable.offset.toInt()))
+    query.maxResults = pageable.pageSize
+
+    val results = query.resultList as List<PersonEntity>
+
+    val countQuery = countMatchCandidates(personQuery, searchCriteria)
```
H-Iwanejko, line 28: I think this wants to be outside of the pageable scope first, otherwise this we be called multiple times increasing processing time

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonBlockingRulesRepository.kt
```diff
@@ -11,10 +14,29 @@ class PersonBlockingRulesRepository {
   @PersistenceContext
   private val entityManager: EntityManager? = null
 
-  fun findMatchCandidates(searchCriteria: PersonSearchCriteria, personQuery: String): List<PersonEntity> {
+  fun findMatchCandidates(searchCriteria: PersonSearchCriteria, personQuery: String, pageable: Pageable, totalResults: Long?): Page<PersonEntity> {
```
H-Iwanejko, line 17: Probably just set the default results to `0L` here instead

H-Iwanejko (approved):

PR #647, opened 2024-10-09
CPR-398 Create Unmerge records

H-Iwanejko:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/MessagingMultiNodeTestBase.kt
```diff
@@ -206,9 +206,11 @@ abstract class MessagingMultiNodeTestBase : IntegrationTestBase() {
     pnc: String?,
     crn: String = randomCRN(),
     cro: String = randomCro(),
+    title: String = randomName(),
     additionalInformation: AdditionalInformation? = null,
     prisonNumber: String = "",
-    prefix: String = randomName(),
```
markreesmoj, line 211: ❤️ Love the continuous improvements in this area

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TelemetryService.kt
```diff
@@ -52,6 +53,20 @@ class TelemetryService(private val telemetryClient: TelemetryClient) {
...
   }
 
+  fun trackEventWithIds(
+    eventType: TelemetryEventType,
+    person: Person,
+    elementMap: Map<EventKeys, String?> = emptyMap(),
+  ) {
+    val identifierMap = mapOf(
+      EventKeys.SOURCE_SYSTEM to person.sourceSystemType.name,
+      EventKeys.DEFENDANT_ID to person.defendantId,
```
markreesmoj, line 63: I'm not sure this is allowed
One bit of PII is OK e.g. CRN, but something like this which would allow a picture of an individual to be put together might not be. Do we really need this?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/PersonService.kt
```diff
@@ -128,49 +120,11 @@ class PersonService(
...
     return searchService.processCandidateRecords(highConfidenceMatches)
   }
 
-  private fun trackEvent(
-    eventType: TelemetryEventType,
-    person: Person,
-    elementMap: Map<EventKeys, String?> = emptyMap(),
-  ) {
-    val identifierMap = mapOf(
-      EventKeys.SOURCE_SYSTEM to person.sourceSystemType.name,
```
markreesmoj, line 166: :-) I see we already had it

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TelemetryService.kt
H-Iwanejko: Yeah, it only will log out one of them as the rest will be null due to the source system ID 👍

markreesmoj (approved):
👍

PR #648, opened 2024-10-10
CPR-79 fix dependency

H-Iwanejko:


build.gradle.kts
```diff
@@ -30,6 +30,7 @@ dependencies {
   implementation("com.microsoft.azure:applicationinsights-logging-logback:2.6.4")
   implementation("org.springframework.cloud:spring-cloud-starter-openfeign:4.1.3")
   implementation("org.springframework.security:spring-security-config:6.3.3")
+  implementation("commons-io:commons-io:2.17.0")
```
markreesmoj, line 33: We do not have a direct dependency on this, I assume, so something else is probably bringing it in?
I assume we are just trying to make sure we get this version as a transitive dependency?

If so, can we do something like the line directly below instead, putting it in a `constraints` block? That makes our dependency on it a bit clearer.

H-Iwanejko: Updated 👍

markreesmoj (approved):
👍

PR #649, opened 2024-10-10
CPR-430-create-event-logging-table

JunaedAhmedMOJ:


src/main/resources/db/migration/V63__add_event_logging_table.sql
```diff
@@ -0,0 +1,17 @@
...
+----------------------------------------
+
+DROP TABLE IF EXISTS event_logging;
+CREATE TABLE IF NOT EXISTS event_logging
+(
+    id                                      UUID        PRIMARY KEY,
+    before_data                             JSONB       DEFAULT NULL,
+    processed_data                          JSONB       DEFAULT NULL,
+    source_system_id                        TEXT        DEFAULT NULL,
+    data source                             TEXT        DEFAULT NULL,
```
H-Iwanejko, line 11: missing an underscore `data_source `

src/main/resources/db/migration/V63__add_event_logging_table.sql
```diff
@@ -0,0 +1,18 @@
...
+----------------------------------------
+
+DROP TABLE IF EXISTS event_logging;
+CREATE TABLE IF NOT EXISTS event_logging
+(
+    id                                      UUID        PRIMARY KEY,
+    before_data                             JSONB       DEFAULT NULL,
+    processed_data                          JSONB       DEFAULT NULL,
+    source_system_id                        TEXT        DEFAULT NULL,
+    UUID                                    UUID        DEFAULT NULL,
```
H-Iwanejko, line 11: might as well keep this lower case as well

H-Iwanejko (approved):

src/main/resources/db/migration/V63__add_event_logging_table.sql
markreesmoj: Is this the personKey? If so should it be a foreign key? Also should it be named person_key?

src/main/resources/db/migration/V63__add_event_logging_table.sql
```diff
@@ -0,0 +1,18 @@
+BEGIN;
+----------------------------------------
+
+DROP TABLE IF EXISTS event_logging;
+CREATE TABLE IF NOT EXISTS event_logging
+(
+    id                                      UUID        PRIMARY KEY,
+    before_data                             JSONB       DEFAULT NULL,
+    processed_data                          JSONB       DEFAULT NULL,
+    source_system_id                        TEXT        DEFAULT NULL,
```
markreesmoj, line 10: Is this actually an ID?
Looks like a string

`source-system` maybe

src/main/resources/db/migration/V63__add_event_logging_table.sql
```diff
@@ -0,0 +1,18 @@
...
+    uuid                                    UUID        DEFAULT NULL,
+    data_source                             TEXT        DEFAULT NULL,
```
markreesmoj, line 12: Not sure what this is but it will probably become clearer when the code which uses it goes in!
You can safely ignore these comments until then I reckon

PR #657, opened 2024-10-15
CPR-457-add-correlation-id

JunaedAhmedMOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TelemetryService.kt
```diff
@@ -5,6 +5,7 @@ import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.personrecord.model.person.Person
 import uk.gov.justice.digital.hmpps.personrecord.service.type.TelemetryEventType
+import java.util.*
```
H-Iwanejko, line 8: Dont need this

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TelemetryService.kt
```diff
@@ -71,7 +72,21 @@ class TelemetryService(private val telemetryClient: TelemetryClient) {
 
   fun trackEvent(eventType: TelemetryEventType, customDimensions: Map<EventKeys, String?>) {
     log.debug("Sending telemetry event ${eventType.eventName} ")
-    val transformedDimensions: Map<String, String?> = customDimensions.entries.associate { it.key.name to it.value }
-    telemetryClient.trackEvent(eventType.eventName, transformedDimensions, null)
+
+    val operationId = getOperationId()
```
H-Iwanejko, line 76: Could simplify this to:
```
val correlationId: String
    get() = telemetryClient.context.operation.id ?: UUID.randomUUID().toString()
```

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TelemetryService.kt
```diff
@@ -71,7 +72,21 @@ class TelemetryService(private val telemetryClient: TelemetryClient) {
...
+
+    val correlationId = if (operationId != "No OperationId") {
```
H-Iwanejko, line 78: See above

H-Iwanejko (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TelemetryService.kt
```diff
@@ -71,7 +73,12 @@ class TelemetryService(private val telemetryClient: TelemetryClient) {
 
   fun trackEvent(eventType: TelemetryEventType, customDimensions: Map<EventKeys, String?>) {
     log.debug("Sending telemetry event ${eventType.eventName} ")
-    val transformedDimensions: Map<String, String?> = customDimensions.entries.associate { it.key.name to it.value }
-    telemetryClient.trackEvent(eventType.eventName, transformedDimensions, null)
+
+    val correlationId: String=  telemetryClient.context.operation.id ?: UUID.randomUUID().toString()
+
+    val updatedDimensions = customDimensions.entries.associate { it.key.name to it.value } + mapOf("CORRELATION_ID" to correlationId)
```
H-Iwanejko, line 79: Probably worth adding: `CORRELATION_ID` to the event keys

H-Iwanejko (commented):
Can we have a test to ensure the correlation ID is logged out

markreesmoj:
Under what circumstances can we not simply use operationId?

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TelemetryServiceIntTest.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import uk.gov.justice.digital.hmpps.personrecord.config.IntegrationTestBase
+
+class TelemetryServiceIntTest : IntegrationTestBase() {
+
+  @Test
+  fun `should log correlation id in telemetry events`() {
+    val telemetryLogs = telemetryRepository.findAll()
```
H-Iwanejko, line 11: This test assumes there is a telemetry log in the repository to begin with. If you ran this in isolation it would fail

H-Iwanejko: Probably best to check if it is added to a event for Probation, Prison event. Then you can check if it is the same throughout

JunaedAhmedMOJ:
> Under what circumstances can we not simply use operationId, which should be available in AppInsights on all operations?

i guess we would still want to generate a correlation id if it is the first event that happens

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/prison/PrisonEventListenerIntTest.kt
```diff
@@ -133,6 +133,21 @@ class PrisonEventListenerIntTest : MessagingMultiNodeTestBase() {
...
+
+    val correlationId = telemetryClient.context.operation.id
+
+    stubPrisonResponse(ApiResponseSetup(prisonNumber = prisonNumber))
+
+    val additionalInformation = AdditionalInformation(prisonNumber = prisonNumber, categoriesChanged = emptyList())
+    val domainEvent = DomainEvent(eventType = PRISONER_CREATED, personReference = null, additionalInformation = additionalInformation)
+    publishDomainEvent(PRISONER_CREATED, domainEvent)
+
+    checkTelemetry(MESSAGE_RECEIVED, mapOf("PRISON_NUMBER" to prisonNumber, "EVENT_TYPE" to PRISONER_CREATED, "SOURCE_SYSTEM" to "NOMIS", "CORRELATION_ID" to correlationId))
```
H-Iwanejko, line 148: This is checking if correlationID is equal to null.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TelemetryService.kt
```diff
@@ -70,8 +66,10 @@ class TelemetryService(private val telemetryClient: TelemetryClient) {
   }
 
   fun trackEvent(eventType: TelemetryEventType, customDimensions: Map<EventKeys, String?>) {
-    log.debug("Sending telemetry event ${eventType.eventName} ")
-    val transformedDimensions: Map<String, String?> = customDimensions.entries.associate { it.key.name to it.value }
-    telemetryClient.trackEvent(eventType.eventName, transformedDimensions, null)
+    val correlationId = telemetryClient.context.operation.id
+
+    val updatedDimensions = customDimensions.entries.associate { it.key.name to it.value } + mapOf("CORRELATION_ID" to correlationId)
```
markreesmoj, line 71: Is there any point in adding this correlation ID if it is already present as the operationId?

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/prison/PrisonEventListenerIntTest.kt
H-Iwanejko: The ID will only be generated when the telemetry client is initialized

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TelemetryService.kt
markreesmoj: This whole change doesn't seem to add any value - we are just logging this value twice instead of once as far as I can see

H-Iwanejko: I agree this doesn't need to be added to the exisiting telemetry. However, it does need pulling out so we can add it it to the audit / event store.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationEventListenerIntTest.kt
```diff
@@ -306,6 +308,19 @@ class ProbationEventListenerIntTest : MessagingMultiNodeTestBase() {
...
+  fun `should log correlation id to telemetry`() {
+    val crn = randomCRN()
+
+    probationDomainEventAndResponseSetup(NEW_OFFENDER_CREATED, ApiResponseSetup(crn = crn))
+
+    val testCorrelationId = (telemetryClient as TelemetryTestConfig.OurTelemetryClient).testCorrelation
+
+    assertNotNull(testCorrelationId)
+
+    checkTelemetry(MESSAGE_RECEIVED, mapOf("CRN" to crn, "EVENT_TYPE" to NEW_OFFENDER_CREATED, "SOURCE_SYSTEM" to "DELIUS"))
```
H-Iwanejko, line 321: Don't need this

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/telemetry/TelemetryTestConfig.kt
```diff
@@ -16,12 +17,19 @@ class TelemetryTestConfig {
...
   class OurTelemetryClient(private val telemetryRepository: TelemetryTestRepository, private val objectMapper: ObjectMapper) : TelemetryClient() {
+    val testCorrelation: String = UUID.randomUUID().toString()
+
     override fun trackEvent(
       event: String?,
       properties: MutableMap<String, String>?,
       metrics: MutableMap<String, Double>?,
     ) {
-      val telemetry = TelemetryEntity(event = event, properties = objectMapper.writeValueAsString(properties))
+      // val correlationId = UUID.randomUUID().toString()
```
H-Iwanejko, line 27: Remove

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/telemetry/TelemetryTestConfig.kt
```diff
@@ -16,12 +17,19 @@ class TelemetryTestConfig {
...
+      val updatedProperties = properties?.toMutableMap() ?: mutableMapOf()
+      updatedProperties["CORRELATION_ID"] = testCorrelation
```
H-Iwanejko, line 29: Don't need to add to properties of telemetry in test

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TelemetryService.kt
H-Iwanejko, line 71: Needs removing as not storing in properties

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TelemetryService.kt
```diff
@@ -10,6 +9,7 @@ enum class EventKeys {
   MESSAGE_ID,
   SOURCE_SYSTEM,
   EVENT_TYPE,
+  CORRELATION_ID,
```
H-Iwanejko, line 12: Can remove

H-Iwanejko (changes requested):

markreesmoj:
I don't think I understand this PR. Is it actually adding any functionality to the application? It doesn't seem like it is

H-Iwanejko (approved):

PR #658, opened 2024-10-16
CPR-465 Create api

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/exceptions/ExceptionHandler.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.http.ResponseEntity
+import org.springframework.security.access.AccessDeniedException
+import org.springframework.security.core.AuthenticationException
+import org.springframework.web.bind.annotation.ExceptionHandler
+import org.springframework.web.bind.annotation.RestControllerAdvice
+
+@RestControllerAdvice
+class ExceptionHandler {
+
+  data class ErrorResponse(
```
markreesmoj, line 13: Consider using this instead perhaps?
https://github.com/ministryofjustice/hmpps-kotlin-lib/blob/main/hmpps-kotlin-spring-boot-autoconfigure/src/main/kotlin/uk/gov/justice/hmpps/kotlin/common/ErrorResponse.kt

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/SearchController.kt
```diff
@@ -0,0 +1,59 @@
...
+import uk.gov.justice.digital.hmpps.personrecord.model.types.SourceSystemType
+
+@RestController
+@RequestMapping("/api")
+@PreAuthorize("hasRole('ROLE_CORE_PERSON_RECORD_API__SEARCH__RO')")
+class SearchController(
+  private val personRepository: PersonRepository,
+) {
+
+  @GetMapping("/search/{personId}")
```
markreesmoj, line 25: Do I understand this correctly - we are providing an endpoint where you can pass any identifier and we will try to match it?
Surely the calling code knows what kind of identifier it is and could tell us? Something like this?
'/search/probation/{personId}' 
This seems like it will make (currently) 2 database calls which will never return anything.
Also, although it seems unlikely, if we end up with an identifier used by more than one source system in the future this will be a problem.

We are bringing complexity into our app by offering this endpoint - is it worth it?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonRepository.kt
```diff
@@ -4,14 +4,20 @@ import org.springframework.data.jpa.repository.JpaRepository
...
+import uk.gov.justice.digital.hmpps.personrecord.model.types.SourceSystemType.DELIUS
 import uk.gov.justice.digital.hmpps.personrecord.model.types.SourceSystemType.NOMIS
 
 @Repository
 interface PersonRepository : JpaRepository<PersonEntity, Long> {
 
   fun findByDefendantId(defendantId: String): PersonEntity?
 
+  fun findByDefendantIdAndSourceSystem(defendantId: String, sourceSystem: SourceSystemType? = COMMON_PLATFORM): PersonEntity?
+
```
markreesmoj, line 17: How does this differ from `findByDefendantId`?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonRepository.kt
```diff
@@ -4,14 +4,20 @@ import org.springframework.data.jpa.repository.JpaRepository
...
   fun findByCrn(crn: String): PersonEntity?
 
+  fun findByCrnAndSourceSystem(crn: String, sourceSystem: SourceSystemType? = DELIUS): PersonEntity?
```
markreesmoj, line 20: and this from `findByCrn`?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/SearchController.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.personrecord.api.controller.exceptions.PersonRecordNotFoundException
+import uk.gov.justice.digital.hmpps.personrecord.api.model.PersonIdentifierRecord
+import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.PersonEntity
+import uk.gov.justice.digital.hmpps.personrecord.jpa.repository.PersonRepository
+import uk.gov.justice.digital.hmpps.personrecord.model.types.SourceSystemType
+
+@RestController
+@RequestMapping("/api")
```
markreesmoj, line 19: It's a minor point but I am not a big fan of splitting the request mapping over multiple anotations.
If we end up having 15 endpoints under `/api` say, the temptation (perhaps even requirement, not sure how the RequestMapping annotation works) would be that they all have to live in this controller, whether they are related or not.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/SearchController.kt
```diff
@@ -0,0 +1,59 @@
...
+    return when {
+      personRecord != PersonEntity.empty -> buildListOfLinkedRecords(personRecord!!)
+      else -> throw PersonRecordNotFoundException(personId)
+    }
+  }
+
+  private fun buildListOfLinkedRecords(personEntity: PersonEntity): List<PersonIdentifierRecord> {
+    return personEntity.personKey?.personEntities?.mapNotNull {
+      buildIdentifierRecord(it)
+    } ?: listOfNotNull(buildIdentifierRecord(personEntity))
```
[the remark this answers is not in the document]
markreesmoj: Looks like we do (maybe for records which are not cmplete enough to match?) - I think we could question this as it looks like it is bringing complexity.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/controller/SearchIntTest.kt
```diff
@@ -0,0 +1,204 @@
...
+  @Test
+  fun `should return NOT FOUND when empty personId`() {
+    webTestClient.get()
+      .uri(searchUrl(""))
+      .authorised(listOf(Roles.ROLE_CORE_PERSON_RECORD_API__SEARCH__RO.name))
+      .exchange()
+      .expectStatus()
+      .isNotFound
+  }
+}
```
markreesmoj, line 204: These tests are very thorough and clear

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/SearchController.kt
H-Iwanejko: Fair point, updated 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonRepository.kt
H-Iwanejko: Removed 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonRepository.kt
H-Iwanejko: Removed 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/SearchController.kt
H-Iwanejko: Discussed and updated 👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/SearchController.kt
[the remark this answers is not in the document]
markreesmoj: Let's get an answer to this question - we might not want to support this.
I can see this endpoint ultimately returning the UUID or whatever we decide to sue as a business key as well

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/SearchController.kt
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  private fun buildIdentifierRecord(personEntity: PersonEntity): PersonIdentifierRecord? {
+    return when (personEntity.sourceSystem) {
+      SourceSystemType.DELIUS -> PersonIdentifierRecord(id = personEntity.crn!!, SourceSystemType.DELIUS.name)
+      SourceSystemType.NOMIS -> PersonIdentifierRecord(id = personEntity.prisonNumber!!, SourceSystemType.NOMIS.name)
+      SourceSystemType.COMMON_PLATFORM -> PersonIdentifierRecord(id = personEntity.defendantId!!, SourceSystemType.COMMON_PLATFORM.name)
+      else -> null
```
markreesmoj, line 51: the null condition will never happen the way this class is now written. Plus we actually know  in the controller method which one of these is the case. I think we can do without this complexity  - maybe this can live in the @GetMapping method or be passed down from it as a function?

H-Iwanejko: This null can happen as this also builds  all the records in the cluster LIBRA etc. not just the records we are looking for.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/SearchController.kt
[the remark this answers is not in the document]
H-Iwanejko: Based on a discussion yesterday, we are returning the record until we rethink not creating a UUID

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/SearchController.kt
markreesmoj: I misunderstood this, sorry

markreesmoj (approved):
👍

PR #659, opened 2024-10-18
CPR-478 Add log events to api

H-Iwanejko:


JunaedAhmedMOJ (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/SearchController.kt
```diff
@@ -10,34 +10,45 @@ import uk.gov.justice.digital.hmpps.personrecord.api.model.PersonIdentifierRecor
...
-    else -> throw PersonRecordNotFoundException(identifier)
+  private fun handlePersonRecord(personEntity: PersonEntity?, identifier: Pair<EventKeys, String>): List<PersonIdentifierRecord> = when {
+    personEntity != PersonEntity.empty -> buildListOfLinkedRecords(personEntity!!, identifier)
+    else -> throw PersonRecordNotFoundException(identifier.second)
   }
 
-  private fun buildListOfLinkedRecords(personEntity: PersonEntity): List<PersonIdentifierRecord> {
+  private fun buildListOfLinkedRecords(personEntity: PersonEntity, identifier: Pair<EventKeys, String>): List<PersonIdentifierRecord> {
+    telemetryService.trackEvent(
+      TelemetryEventType.CPR_SEARCH_REQUEST,
```
markreesmoj, line 46: What does this give us over a simple appInsights query on `requests`?
I'm not sure this adds any value

H-Iwanejko: Only value is seeing what UUID is returned as part of the request.

H-Iwanejko: In the case that we dont want to be raising any excess app insights events. I'll close the PR and we can just query to find the UUID from the record

markreesmoj: if it's useful to be able to find out what UUID is returned then fair enough

PR #665, opened 2024-10-23
CPR-445 Create delete listener and use composition

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/format/EncodingService.kt
```diff
@@ -1,28 +1,24 @@
...
+import uk.gov.justice.digital.hmpps.personrecord.client.model.prisoner.Prisoner
 import uk.gov.justice.digital.hmpps.personrecord.service.RetryExecutor
 
-@Component
-class BaseProbationEventProcessor {
-
-  @Autowired
-  private lateinit var corePersonRecordAndDeliusClient: CorePersonRecordAndDeliusClient
+@Service
+class EncodingService(
```
H-Iwanejko, line 14: Better name for this welcomed!

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationDeletionEventListener.kt
```diff
@@ -0,0 +1,71 @@
...
+import uk.gov.justice.digital.hmpps.personrecord.service.EventKeys.MESSAGE_ID
+import uk.gov.justice.digital.hmpps.personrecord.service.EventKeys.SOURCE_SYSTEM
+import uk.gov.justice.digital.hmpps.personrecord.service.TelemetryService
+import uk.gov.justice.digital.hmpps.personrecord.service.type.OFFENDER_GDPR_DELETION
+import uk.gov.justice.digital.hmpps.personrecord.service.type.TelemetryEventType.MESSAGE_PROCESSING_FAILED
+
+const val PROBATION_DELETION_EVENT_QUEUE_CONFIG_KEY = "cprdeliusdeleteeventsqueue"
+
+@Component
+@Profile("!seeding")
```
markreesmoj, line 25: This is an accident waiting to happen! Not sure what we can do that is any better though

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationDeletionEventListener.kt
```diff
@@ -0,0 +1,71 @@
...
+        log.info("Received a message I wasn't expecting Type: ${sqsMessage.type}")
+      }
+    }
+  }
+
+  private fun handleDeleteEvent(domainEvent: DomainEvent, messageId: String?) {
+    try {
+      val crn = domainEvent.personReference?.identifiers?.first { it.type == "CRN" }!!.value
+      probationDeleteProcessor.processEvent(crn, domainEvent.eventType)
+    } catch (e: FeignException.NotFound) {
```
markreesmoj, line 57: Curious to know if this exception will ever be thrown. It isn't at the moment as far as I can tell. Will deletion involve an HTTP call?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/format/EncodingService.kt
```diff
@@ -1,28 +1,24 @@
...
 import uk.gov.justice.digital.hmpps.personrecord.client.model.offender.ProbationCase
+import uk.gov.justice.digital.hmpps.personrecord.client.model.prisoner.Prisoner
 import uk.gov.justice.digital.hmpps.personrecord.service.RetryExecutor
 
-@Component
-class BaseProbationEventProcessor {
-
-  @Autowired
-  private lateinit var corePersonRecordAndDeliusClient: CorePersonRecordAndDeliusClient
+@Service
```
markreesmoj, line 13: I know it doesn't matter, but this annotation does nothing that @Component doesn't do. It feels like some idea that someone had about 20 years ago that got as far as the language spec and no further :-)

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/format/EncodingService.kt
markreesmoj: I think it's hard to name because it does 2 things. It would be something like a `PrisonerSearchService` or a `ProbationSearchService`
What do we gain by having these 2 methods in this class?

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/MessagingMultiNodeTestBase.kt
```diff
@@ -73,6 +73,10 @@ abstract class MessagingMultiNodeTestBase : IntegrationTestBase() {
     hmppsQueueService.findByQueueId("cprdeliusmergeeventsqueue")
   }
 
+  val probationDeleteEventsQueue by lazy {
+    hmppsQueueService.findByQueueId("cprdeliusdeleteeventsqueue")
+  }
```
markreesmoj, line 78: It will probably be worthwhile to purge this queue before running any tests - I think we do that with all the others

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationEventListenerIntTest.kt
```diff
@@ -61,7 +61,7 @@ class ProbationEventListenerIntTest : MessagingMultiNodeTestBase() {
     telemetryRepository.deleteAll()
   }
 
-  fun probationUrl(crn: String) = "/probation-cases/$crn"
+  private fun probationUrl(crn: String) = "/probation-cases/$crn"
```
markreesmoj, line 64: ❤️

markreesmoj (approved):
👍 
I think this is fine. It is a big change to review  so it's hard to be sure.
Maybe because it feels like 3 changes - a package reorganisation, the removal of the subclassing and the introduction of the deletion event handling

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/MessagingMultiNodeTestBase.kt
H-Iwanejko: Will update in further PR's

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationDeletionEventListener.kt
H-Iwanejko: Good point 👍

PR #688, opened 2024-11-19
CPR-503 consume recluster events

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/person/ReclusterService.kt
```diff
@@ -62,7 +57,6 @@ class ReclusterService(
   private fun handleSingleRecordInCluster(personKeyEntity: PersonKeyEntity) {
     val record = personKeyEntity.personEntities.first()
     val highConfidenceMatches = searchService.findCandidateRecordsWithUuid(record)
-      .filterNot { it.candidateRecord.id == record.id }
```
H-Iwanejko, line 65: No longer need this as the own record is filtered at DB query level

markreesmoj (approved):
👍

PR #698, opened 2024-11-27
CPR-500 change dlq max recieve

H-Iwanejko:


Makefile
```diff
@@ -17,7 +17,8 @@ endif
 stop-containers:
 	docker compose down
 
-restart-containers: stop-containers start-containers
+restart-containers: stop-containers
+	docker compose up -d
```
H-Iwanejko, line 21: manual start up instead of check as brief peroid of still `up`

src/test/resources/application-test.yml
```diff
@@ -45,29 +46,35 @@ hmpps.sqs:
       dlqName: cpr_delius_offender_events_queue_dlq
       subscribeTopicId: domainevents
       subscribeFilter: '{"eventType":[ "probation-case.engagement.created", "OFFENDER_ADDRESS_CHANGED", "OFFENDER_DETAILS_CHANGED", "OFFENDER_ALIAS_CHANGED"] }'
+      dlqMaxReceiveCount: 1
```
H-Iwanejko, line 49: For envs this is set to 3. But I believe we dont need to retry on failures if forced as we should expect them within tests.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationMergeEventListenerIntTest.kt
```diff
@@ -335,16 +334,14 @@ class ProbationMergeEventListenerIntTest : MessagingMultiNodeTestBase() {
       ),
     )
 
-    probationMergeEventsQueue!!.sqsClient.purgeQueue(
```
H-Iwanejko, line 338: As the queues dlqMaxReceiveCount were set to 3 purging the queue after the first would cause the SQS error locally. However, still unsure of why this occurred in prod

src/test/resources/application-test.yml
H-Iwanejko: I could set to 3 but would require to update tests to expect 3 log statements for message recieved?

PR #700, opened 2024-11-28
CPR-500 Add timeout to listeners

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/court/CourtEventListener.kt
```diff
@@ -14,31 +15,29 @@ import uk.gov.justice.digital.hmpps.personrecord.service.EventKeys.EVENT_TYPE
...
 @Component
 @Profile("!seeding")
 class CourtEventListener(
-  val objectMapper: ObjectMapper,
-  val courtEventProcessor: CourtEventProcessor,
-  val telemetryService: TelemetryService,
+  private val objectMapper: ObjectMapper,
+  private val courtEventProcessor: CourtEventProcessor,
+  private val telemetryService: TelemetryService,
+  @Value("\${timeout.message}") private val messageTimeoutMs: Long = 90000,
```
H-Iwanejko, line 29: Just set the timeout to an arbitrary 90 secs, as the message visibility is 120

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TimeoutExecutor.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.justice.digital.hmpps.personrecord.service
+
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
+
+object TimeoutExecutor {
+
+  fun runWithTimeout(timeout: Long, task: () -> Unit) = runBlocking {
+    withTimeout(timeout) {
```
markreesmoj, line 9: maybe just hard code this for now rather than wire the same value in everywhere and pass it through - would make this PR a lot smaller

src/test/resources/application-test.yml
```diff
@@ -23,6 +23,9 @@ core-person-record-and-delius:
 retry:
   delay: 450
 
+timeout:
```
markreesmoj, line 26: if we're not going to change it, is there any need to specify it again?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/court/CourtEventTempListener.kt
```diff
@@ -25,15 +27,14 @@ const val CPR_COURT_EVENTS_TEMP_QUEUE_CONFIG_KEY = "cprcourtcaseeventstemporaryq
...
+  private val telemetryService: TelemetryService,
+  private val queueService: QueueService,
+  @Value("\${timeout.message}") private val messageTimeoutMs: Long = 90000,
 ) {
 
   @SqsListener(CPR_COURT_EVENTS_TEMP_QUEUE_CONFIG_KEY, factory = "hmppsQueueContainerFactoryProxy")
-  fun onMessage(
-    rawMessage: String,
-  ) {
+  fun onMessage(rawMessage: String) = TimeoutExecutor.runWithTimeout(messageTimeoutMs) {
```
markreesmoj, line 37: I wouldn't bother changing this - I'm going to delete it shortly

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/court/CourtEventFIFOListener.kt
```diff
@@ -19,25 +19,26 @@ import uk.gov.justice.digital.hmpps.personrecord.service.EventKeys.MESSAGE_ID
...
+  private val messageTimeoutMs: Long = 90000,
   @Value("\${retry.delay}")
   private val retryDelay: Long = 0,
 ) {
+
   @SqsListener(CPR_COURT_EVENTS_FIFO_QUEUE_CONFIG_KEY, factory = "hmppsQueueContainerFactoryProxy")
-  fun onMessage(
-    rawMessage: String,
-  ) {
+  fun onMessage(rawMessage: String) = TimeoutExecutor.runWithTimeout(messageTimeoutMs) {
```
markreesmoj, line 40: I wouldn't bother changing this - I'm going to delete it shortly

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TimeoutExecutor.kt
markreesmoj: Can we get a test in for this somehow?
AIUI this will throw an exception so the message processing will fail.
:-)
Maybe set the test timeout to 3s or something and have an operation delay for 4s, then assert that the message being consumed ends up back on the queue?

markreesmoj (approved):
👍

PR #705, opened 2024-12-02
FIX: upgrade commons-io which is brought in by feign

markreesmoj:


build.gradle.kts
```diff
@@ -34,15 +34,11 @@ dependencies {
...
+    implementation("commons-io:commons-io:2.18.0") {
       because("2.13.0 has CVEs")
     }
   }
-  constraints {
-    implementation("org.bouncycastle:bcprov-jdk18on:1.79") {
-      because("1.77 has CVEs")
-    }
-  }
+
```
markreesmoj, line 41: bouncycastle seems to be on 1.78 now (through feign) which does not have a CVE

JunaedAhmedMOJ (approved):

PR #706, opened 2024-12-02
Speed up tests

markreesmoj:


src/main/resources/logback-spring.xml
```diff
@@ -14,22 +14,6 @@
         <appender name="aiAppender" class="com.microsoft.applicationinsights.logback.ApplicationInsightsAppender"/>
     </springProfile>
 
-    <springProfile name="!log-json">
```
markreesmoj, line 17: Unused I think

build.gradle.kts
```diff
@@ -81,7 +81,11 @@ tasks {
   register("initialiseDatabase", Test::class) {
     include("**/HealthCheckIntTest.class")
   }
-
+  register("fastTest", Test::class) {
```
markreesmoj, line 84: Just added this to pinpoint that the problem is actually the listener tests

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/MessagingMultiNodeTestBase.kt
```diff
@@ -352,44 +352,6 @@ abstract class MessagingMultiNodeTestBase : IntegrationTestBase() {
...
-    await.atMost(Duration.ofSeconds(2)) untilCallTo {
-      prisonMergeEventsQueue!!.sqsClient.countAllMessagesOnQueue(prisonMergeEventsQueue!!.queueUrl).get()
-    } matches { it == 0 }
-    await.atMost(Duration.ofSeconds(2)) untilCallTo {
-      prisonMergeEventsQueue!!.sqsDlqClient!!.countAllMessagesOnQueue(prisonMergeEventsQueue!!.dlqUrl!!).get()
-    } matches { it == 0 }
-
-    await.atMost(Duration.ofSeconds(2)) untilCallTo {
-      reclusterEventsQueue!!.sqsClient.countAllMessagesOnQueue(reclusterEventsQueue!!.queueUrl).get()
-    } matches { it == 0 }
```
markreesmoj, line 392: It's probably OK if we just check the last one  we purged is empty - all the others should be empty by this point

markreesmoj:
Seems to knock a couple of minutes off
[Some nice green arrows](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-person-record?branch=speed-up-tests)

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/MessagingMultiNodeTestBase.kt
H-Iwanejko: I know its inheritance again in another level, but could we do something like e.g. `UnmergeMessageingMultiNodeBase` where we only purge the queues we need to rather than all of them 🤔

H-Iwanejko (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/MessagingMultiNodeTestBase.kt
markreesmoj: 🤔 I don't mind inheritance in tests if it speeds things up.

PR #707, opened 2024-12-02
CPR-498 Add insight appender for preprod and dev

H-Iwanejko:


src/main/resources/logback-spring.xml
```diff
@@ -10,7 +10,7 @@
         <appender name="aiAppender" class="ch.qos.logback.core.helpers.NOPAppender"/>
     </springProfile>
 
-    <springProfile name="prod">
+    <springProfile name="!local">
```
markreesmoj, line 13: doesn't this conflict with lines 9-11?
[I don't actually think we need this appender at all](https://mojdt.slack.com/archives/C012FNNQ4A3/p1677234268902449?thread_ts=1677233333.667589&cid=C012FNNQ4A3).
Let's take it out and see what happens

markreesmoj (commented):
💬

src/main/resources/logback-spring.xml
H-Iwanejko: Removed 👍

markreesmoj (approved):
👍

PR #715, opened 2024-12-03
CPR-486 increase memory capacity

JunaedAhmedMOJ:
due to out of memory errors, the theory is that the heap space is being exceeded (heap space is the xmx setting)

in order to increase the heap space we need to also increase the limit, this can be observed by the link below :

https://grafana.live.cloud-platform.service.justice.gov.uk/d/6581e46e4e5c7ba40a07646395ef7b23/kubernetes-compute-resources-pod?var-datasource=prometheus&var-cluster=&var-namespace=hmpps-person-record-prod&var-pod=hmpps-person-record-c449f97f7-ftwwq&from=now-2d&to=now&timezone=utc&refresh=10s


H-Iwanejko (changes requested):
Failing tests

PR #716, opened 2024-12-04
CPR-79 Delete in batch

H-Iwanejko:
* Should resolve memory issue for term frequency


markreesmoj:
worth a try, maybe bring forward the job runtime in preprod to 1115 or so to test

markreesmoj (approved):

PR #717, opened 2024-12-04
FIX: do not run initialiseDatabase task as part of build - only on de…

markreesmoj:
…mand

Should knock about 45 seconds off the build in CI

`./gradlew initialiseDatabase' still works and it the only way to run that task

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -80,6 +80,10 @@ tasks {
     exclude("**/InitialiseDatabase.class")
   }
 
+  getByName("initialiseDatabase") {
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -80,6 +80,10 @@ tasks {
     exclude("**/InitialiseDatabase.class")
   }
 
+  getByName("initialiseDatabase") {
+    onlyIf { gradle.startParameter.taskNames.contains("initialiseDatabase") }
+  }
+
```
markreesmoj, line 86: `> Task :initialiseDatabase SKIPPED

> Task :test
`

markreesmoj: ❤️

PR #723, opened 2024-12-05
CPR-510 Person Aggregate Data Table

H-Iwanejko:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/jobs/GenerateTermFrequenciesIntTest.kt
```diff
@@ -64,17 +63,15 @@ class GenerateTermFrequenciesIntTest : WebTestBase() {
...
-      PersonEntity.from(
-        Person.from(
-          Defendant(
-            pncId = PNCIdentifier.from(pnc),
-            personDefendant = PersonDefendant(
-              personDetails = PersonDetails(
-                firstName = randomName(),
-                lastName = randomName(),
-                gender = "Male",
-              ),
```
markreesmoj, line 77: ❤️

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/termfrequency/TermCleaningRepository.kt
```diff
@@ -0,0 +1,92 @@
...
+
+  @Transactional
+  fun deletePersonAggregateData() = entityManager.createNativeQuery(DELETE_PERSON_AGGREGATE_SQL).executeUpdate()
+
+  @Transactional
+  fun collectPersonAggregateData() = entityManager.createNativeQuery(COLLECT_PERSON_AGGREGATE_SQL).executeUpdate()
+
+  companion object {
+    private const val COLLECT_PERSON_AGGREGATE_SQL = """
+      CREATE MATERIALIZED VIEW personrecordservice.person_aggregate_data as (
```
markreesmoj, line 20: is this a good idea?
What about anything which depends on the view running while this job is running?
Should we have this defined in flyway instead and refresh it?
Maybe one for the 11am meeting?

markreesmoj: The impression I got from the meeting today was that it would be better to refresh the view rather than drop and recreate it. This would also give us the advantage of the view appearing in a database schema report (do we do one for this schema 🤔 we should do). It would also be better to have the CREATE VIEW code in Flyway so we can find it easily

src/main/resources/db/migration/V69__create_person_aggregate_materialized_view.sql
```diff
@@ -0,0 +1,85 @@
...
+    LEFT JOIN
+        address_agg addr ON p.id = addr.fk_person_id
+    LEFT JOIN
+        reference_agg r ON p.id = r.fk_person_id
+    LEFT JOIN
+        sentence_info_agg s ON p.id = s.fk_person_id
+    ORDER BY p.id
+);
+
+CREATE UNIQUE INDEX idx_person_aggregate_data_id ON personrecordservice.person_aggregate_data (id);
```
H-Iwanejko, line 82: Requires a unique index to created on materialised view to be able to concurrently refresh the table

markreesmoj (approved):
👍

PR #726, opened 2024-12-06
FIX: merge can be to either candidate so take account of this in test

markreesmoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/cpr/ReclusterEventListenerIntTest.kt
```diff
@@ -370,19 +370,23 @@ class ReclusterEventListenerIntTest : MessagingMultiNodeTestBase() {
...
     awaitAssert {
-      val sourceCluster = personKeyRepository.findByPersonId(cluster1.personId)
-      assertThat(sourceCluster?.status).isEqualTo(UUIDStatusType.RECLUSTER_MERGE)
-      assertThat(sourceCluster?.personEntities?.size).isEqualTo(0)
-      assertThat(sourceCluster?.mergedTo).isEqualTo(cluster2.id)
+      val mergeSource = personKeyRepository.findByPersonId(cluster1.personId)
+      assertThat(mergeSource?.status).isEqualTo(UUIDStatusType.RECLUSTER_MERGE)
+      assertThat(mergeSource?.personEntities?.size).isEqualTo(0)
+    }
+    val sourceCluster = personKeyRepository.findByPersonId(cluster1.personId)
```
markreesmoj, line 377: As far as I can tell there is no saying whether cluster1 will be merged to cluster2 or cluster3. The frequent test failures seem to indicate this. The code picks the first [high confidence match](https://github.com/ministryofjustice/hmpps-person-record/blob/cf6438f05fb37d1a461de66c5ca7371c89667703/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/ReclusterService.kt#L70)

markreesmoj: [The records are ordered](https://github.com/ministryofjustice/hmpps-person-record/blob/e9e82cf61f003bff87caa2af3b00323205d80fa1/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/search/SearchService.kt#L94) but we give the same match probability to both.

Another approach to fixing this would be to stub different (high confidence) match scores for each cluster.
In fact that might be better 🤔

H-Iwanejko: I think already tried this, but same behaviour  as still unknown order that it is returned in the database😢

H-Iwanejko (approved):

PR #742, opened 2024-12-11
CPR-79 wait for previous unmerge to finish

markreesmoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationUnmergeEventListenerIntTest.kt
```diff
@@ -427,6 +427,14 @@ class ProbationUnmergeEventListenerIntTest : MessagingMultiNodeTestBase() {
...
+      CPR_RECORD_UNMERGED,
+      mapOf(
+        "REACTIVATED_CRN" to reactivatedCrn,
+        "UNMERGED_CRN" to unmergedCrn,
+        "SOURCE_SYSTEM" to "DELIUS",
+      ),
+    )
   }
 
   @Test
```
markreesmoj, line 440: This test fails frequently because the previous test hasn't finished processing the unmerge event yet, so it makes http calls which are no longer stubbed. The above telemetry check should sort it out

src/test/resources/application-test.yml
```diff
@@ -21,7 +21,7 @@ core-person-record-and-delius:
   base-url: http://localhost:8090
 
 retry:
-  delay: 450
+  delay: 150
```
markreesmoj, line 24: why not? Let's see how reliable this is. Should save a second here and there

H-Iwanejko (approved):
👍

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/prison/PrisonEventListenerIntTest.kt
```diff
@@ -323,6 +324,14 @@ class PrisonEventListenerIntTest : MessagingMultiNodeTestBase() {
...
     val additionalInformation = AdditionalInformation(prisonNumber = prisonNumber)
     val domainEvent = DomainEvent(eventType = PRISONER_CREATED, personReference = null, additionalInformation = additionalInformation)
+    stubMatchScore(
+      MatchResponse(
+        matchProbabilities = mutableMapOf(
+          "0" to 0.9999999,
+          "1" to 0.9999991,
+        ),
+      ),
+    )
```
markreesmoj, line 334: sometimes, but not always, match is called here.
https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-person-record/3770/workflows/f72191dc-53e9-461b-8353-c2218776e6f3/jobs/11575/tests

markreesmoj: Is this a good idea? Might it be better to investigate why match is called at all?

H-Iwanejko: It should try match on create right (if theres candidates)? might be matching on the other records on something in this: `probationDomainEventAndResponseSetup` where similar name etc.

PR #744, opened 2024-12-12
CPR-519-add-indexes

JunaedAhmedMOJ:


src/main/resources/db/migration/V70__add_indexes_for_child_tables.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+-------------------------------------------------------
+CREATE INDEX idx_identifier_type ON reference(identifier_type);
```
markreesmoj, line 3: The convention for our index names (which I think have to be unique in the database) is to use the table name

So these should `be idx_reference_identifer_type`

markreesmoj (commented):
Please rename the indexes to include the table name

markreesmoj (approved):
👍

H-Iwanejko (commented):
Just a question, should these be a combine index of type and value?

JunaedAhmedMOJ:
> Just a question, should these be a combine index of type and value?

hmm i guess it would depend on if we always try and get both values from our query?

PR #747, opened 2024-12-12
CPR-510 update person aggregate data view (2nd Attempt)

H-Iwanejko:
* Now filters out postcodes that are <= 3 so no negative indices are produced


markreesmoj (approved):
🫵

H-Iwanejko:
> is no 71 correct here?

Its next number in the migration, as juneads index addition was 70, right?

markreesmoj (approved):
👍

PR #750, opened 2024-12-12
Revert "CPR-510 update person aggregate data view (3rd Attempt)"

H-Iwanejko:
Reverts ministryofjustice/hmpps-person-record#749
* Query takes too long to allow application start up, needs to run with `WITH NO DATA`


markreesmoj (approved):
👍

PR #755, opened 2024-12-13
CPR-524 Use prepared statements when querying DB

H-Iwanejko:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/IntegrationTestBase.kt
```diff
@@ -67,6 +68,11 @@ class IntegrationTestBase {
 
   fun probationUrl(crn: String) = "/probation-cases/$crn"
 
+  @BeforeEach
+  fun beforeEachIntTest() {
```
H-Iwanejko, line 72: Moved to integration test base as searchIntTest require this to succeed locally on multiple runs

markreesmoj (approved):
Don't fully understand this but it seems reasonable

PR #771, opened 2025-01-09
CPR-512-add-transactional

JunaedAhmedMOJ:
add transactional to CreateUpdateService
remove ReadWriteLockService.kt


H-Iwanejko (approved):
🧹  

This problem which i've seen before usually occurs when we update / create a entity. But the memory entity version is not reflective of the change, which throws the error.

e.g. update address postcode to 'xxx' results in entity version 1. but the entity store in the variable to 'save' is still holding the entity version 0.

markreesmoj:
It seems quite likely that the root cause here is 2 messages in close succession for the same person. We can investigate further if (as we expect) the error still occurs following this change

PR #780, opened 2025-01-13
CPR-512 add a logging line to show the class of the uncaught exception

markreesmoj:
It's hard to tell from the stacktrace which error is being caught in here, so if we log it that should allow us to add it to the list of retryable entity exceptions


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/RetryExecutor.kt
```diff
@@ -48,7 +48,10 @@ object RetryExecutor {
 
             currentDelay = min((currentDelay * EXPONENTIAL_FACTOR).toLong(), maxDelayMillis)
           }
-          else -> throw e
+          else -> {
+            println("Failed to retry on class" + e::class)
```
H-Iwanejko, line 52: Shall we log that out to log.info?

JunaedAhmedMOJ (dismissed):

H-Iwanejko (approved):

PR #800, opened 2025-01-21
CPR-566 reduce use of prefix in names to help with test flakiness

markreesmoj:
We believe that because prefix is rarely used, the first name and last name are quite commonly set to "firstName" and "lastName".
This means that sometimes an unexpected number of match candidates are found and so more recluster events happen than are expected.
The fact is that when this happens it is almost always expecting 1 and getting 5.

We have started by removing the prefix from probation responses when processing messages. We will do another PR doing the same for prison responses which will hopefully further cut the number of flaky tests.


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/test/responses/ProbationCaseResponses.kt
```diff
@@ -27,9 +27,9 @@ fun probationCaseResponse(probationCase: ApiResponseSetup) = """
       "aliases": [
         {
           "name": {
-            "forename": "${probationCase.prefix ?: ""}FirstName",
+            "forename": "AliasFirstName",
```
H-Iwanejko, line 30: Should these be dynamically generated as well? going to start sending them for matching.

markreesmoj: they should really, we were just being lazy. We wondered about swapping lastname and firstname to save on setup but that is probably too lazy as well

H-Iwanejko (approved):

PR #801, opened 2025-01-21
CPR-566-pause-messages

JunaedAhmedMOJ:
pause messaging
alter readiness

### DIFF
--- a/helm_deploy/hmpps-person-record/values.yaml
+++ b/helm_deploy/hmpps-person-record/values.yaml
@@ -1,4 +1,9 @@
 generic-service:
+
+  readinessProbe:
+    periodSeconds: 35
+    failureThreshold: 20
+
   nameOverride: hmpps-person-record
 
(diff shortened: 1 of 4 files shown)


helm_deploy/hmpps-person-record/values.yaml
```diff
@@ -1,4 +1,9 @@
 generic-service:
+
+  readinessProbe:
+    periodSeconds: 35
+    failureThreshold: 20
```
H-Iwanejko, line 5: Assumes this fails after 10 minutes of polling?

markreesmoj: 30 seconds before it starts (default value, not overridden here)
10 second timeout X 20 = 200 seconds (default value, not overridden here)
35 seconds between each X 19 = 665 seconds

915 seconds in all I think which is 15 mins 15 seconds

H-Iwanejko (approved):

PR #813, opened 2025-01-31
CPR-578 Retrieve large messages from S3 for courts

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/SQSMessage.kt
```diff
@@ -18,17 +18,14 @@ data class SQSMessage(
   val messageAttributes: MessageAttributes? = null,
 ) {
   fun getMessageType(): String? = messageAttributes?.messageType?.value
+  fun getEventType(): String? = messageAttributes?.eventType?.value
```
markreesmoj, line 21: This always has the value of messageType instead - can't see why

Obsiye (dismissed):
Awesome, looks good to me

Obsiye (dismissed):

H-Iwanejko (approved):
✅ LGTM

Obsiye (approved):

PR #817, opened 2025-02-06
Generate-accurate-DL

JunaedAhmedMOJ:


H-Iwanejko (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/test/RandomTestData.kt
```diff
@@ -30,7 +30,16 @@ fun randomNationality(): String = randomLowerCaseString()
...
 
-fun randomDriverLicenseNumber(): String = randomLowerCaseString(5).uppercase() + randomDigit(6) + randomLowerCaseString(5).uppercase()
+fun randomDriverLicenseNumber(): String {
+  fun randomString(length: Int, source: String) = (1..length).map { source.random() }.joinToString("")
+  val surnamePart = randomLowerCaseString(5).uppercase()
+  val year = (50..99).random().toString() // Last two digits of birth year (1950-1999)
+  val month = (1..12).random().toString().padStart(2, '0')
+  val day = (1..31).random().toString().padStart(2, '0')
+  val initials = randomLowerCaseString(2).uppercase()
+  val checkDigits = randomString(2, randomDigit(2))
```
markreesmoj, line 40: Could this be `randomDigit(2).toString()`

markreesmoj: Then we would not need the randomString internal function

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/test/RandomTestData.kt
```diff
@@ -30,7 +30,16 @@ fun randomNationality(): String = randomLowerCaseString()
 
 fun randomReligion(): String = randomLowerCaseString()
 
-fun randomDriverLicenseNumber(): String = randomLowerCaseString(5).uppercase() + randomDigit(6) + randomLowerCaseString(5).uppercase()
+fun randomDriverLicenseNumber(): String {
+  fun randomString(length: Int, source: String) = (1..length).map { source.random() }.joinToString("")
+  val surnamePart = randomLowerCaseString(5).uppercase()
+  val year = (50..99).random().toString() // Last two digits of birth year (1950-1999)
+  val month = (1..12).random().toString().padStart(2, '0')
+  val day = (1..31).random().toString().padStart(2, '0')
```
markreesmoj, line 38: It's not a huge deal of course as we don't validate it, but this will not always give a valid date

We could instead use `randomDate().year.toString().takeLast(2)`

The disadvantage here of course is that randomDate returns any date in the last 75 years so it will give birthdates for people who are too young to drive. I don't think this is worth worrying about, though we could modify it to take an end year if we wanted to.

PR #826, opened 2025-02-12
Cpr 582 remove search for libra

JunaedAhmedMOJ:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/SearchServiceIntTest.kt
```diff
@@ -119,70 +68,6 @@ class SearchServiceIntTest : IntegrationTestBase() {
     assertThat(candidateRecords.size).isEqualTo(3)
   }
 
-  @Test
-  fun `should find candidate records on exact matches on PNC`() {
```
H-Iwanejko, line 123: Need to verify that some of these test are still needed but should use the `searchByAllSourceSystemsAndHasUuid ` funtion instead. I think both was used as they were essential the same underneath (relatively). But some of these test still stand... for now.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/SearchServiceIntTest.kt
```diff
@@ -628,95 +129,6 @@ class SearchServiceIntTest : IntegrationTestBase() {
     noCandidatesFound(candidateRecords)
   }
 
-  @Test
-  fun `should order multiple matches descending`() {
```
H-Iwanejko, line 632: See above. Still applies

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/SearchServiceIntTest.kt
```diff
@@ -226,390 +111,6 @@ class SearchServiceIntTest : IntegrationTestBase() {
...
-      lastName = lastName,
-      sourceSystem = COMMON_PLATFORM,
-    )
-    val candidateRecords = searchService.findCandidateRecordsBySourceSystem(searchingPerson)
-
-    noCandidatesFound(candidateRecords)
-  }
-
-  @Test
-  fun `should not find candidate records when record is marked as merged to another record`() {
```
H-Iwanejko, line 571: See above

PR #832, opened 2025-02-13
CPR-565 Improvements to courts processing

H-Iwanejko:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/court/commonplatform/CommonPlatformCourtEventListenerIntTest.kt
```diff
@@ -332,37 +332,38 @@ class CommonPlatformCourtEventListenerIntTest : MessagingMultiNodeTestBase() {
...
         ),
       ),
     )
 
-    checkTelemetry(
-      MESSAGE_RECEIVED,
-      mapOf("MESSAGE_ID" to messageId, "SOURCE_SYSTEM" to COMMON_PLATFORM.name, "EVENT_TYPE" to COMMON_PLATFORM_HEARING.name),
-      times = 0,
-    )
+    assertThat(personRepository.findByDefendantId(defendantId)).isNull()
```
markreesmoj, line 367: This will always be true unless the message processing is ⏩ ⏩ 
`awaitAssert` is useful here.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/court/CourtEventProcessor.kt
```diff
@@ -53,15 +53,13 @@ class CourtEventProcessor(
...
       .flatMap { it.defendants }
       .filterNot { it.isYouth }
+      .filter { it.isPerson() }
       .distinctBy { it.id }
-    val defendantIDs = uniqueDefendants.joinToString(" ") { it.id.toString() }
-    log.debug("Processing Common Platform Event with ${uniqueDefendants.size} distinct defendants with defendantId $defendantIDs")
 
-    val personDefendants = uniqueDefendants.filter { it.isPerson() }
-    personDefendants.forEach { defendant ->
+    uniquePersonDefendants.forEach { defendant ->
```
markreesmoj, line 62: could even do away with this variable and put the .foreach on the end of the above
:-) there is a lot to go at here.
For example, the `CommonPlatformHearingEvent` should actually just be a list of prosecutionCases as far as I can see

PR #833, opened 2025-02-13
Cpr 605 createdat lastmodifiedat timestamps

JunaedAhmedMOJ:


src/main/resources/db/migration/V77__add_createdat_lastmodifiedat_timestamps.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
+-------------------------------------------------------
+
+
+ALTER TABLE person
+     ADD COLUMN created_at TIMESTAMP DEFAULT NULL;
+
+ALTER TABLE person
+     ADD COLUMN last_modified_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP;
```
markreesmoj, line 9: I think we are probably going to want an index on this column.

markreesmoj (approved):
👍

PR #834, opened 2025-02-13
CPR-605-index-timezone

JunaedAhmedMOJ:
create index for timezone

### DIFF
--- a/src/main/resources/db/migration/V78__index_timestamp.sql
+++ b/src/main/resources/db/migration/V78__index_timestamp.sql
@@ -1,0 +1,22 @@
+BEGIN;
+-------------------------------------------------------
+ALTER TABLE IF EXISTS person DROP COLUMN created_at;
+ALTER TABLE IF EXISTS person DROP COLUMN last_modified_at;
+
+ALTER TABLE IF EXISTS person
+     ADD COLUMN created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP;
+
+ALTER TABLE IF EXISTS person
+     ADD COLUMN last_modified_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP;
+
+
+CREATE INDEX idx_person_created_at ON person(created_at);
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/V78__index_timestamp.sql
```diff
@@ -0,0 +1,22 @@
+BEGIN;
+-------------------------------------------------------
+ALTER TABLE IF EXISTS person DROP COLUMN created_at;
+ALTER TABLE IF EXISTS person DROP COLUMN last_modified_at;
+
+ALTER TABLE IF EXISTS person
+     ADD COLUMN created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP;
```
markreesmoj, line 7: I thought we were going to leave this as null? If we have decided otherwise fair enough

markreesmoj: Would also be inclined to drop the _at from the end of the name

src/main/resources/db/migration/V78__index_timestamp.sql
```diff
@@ -0,0 +1,22 @@
...
+
+ALTER TABLE IF EXISTS person
+     ADD COLUMN last_modified_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP;
+
+
+CREATE INDEX idx_person_created_at ON person(created_at);
```
H-Iwanejko, line 13: Actually, I dont think we need an index on `created_at` for now as i dont see us searching on it any time soon

markreesmoj (commented):
👍

src/main/resources/db/migration/V78__index_timestamp.sql
JunaedAhmedMOJ: yep should be null il edit it now

H-Iwanejko: Yeah just seen this, needs to be null 👍

markreesmoj (approved):
👍

PR #847, opened 2025-02-20
CPR-508-canonical-record-api

JunaedAhmedMOJ:
CPR-602 created entity and mapping for raw pnc cro


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/court/commonplatform/Defendant.kt
```diff
@@ -25,6 +25,8 @@ data class Defendant(
   val ethnicity: Ethnicity? = null,
   val aliases: List<DefendantAlias>? = emptyList(),
   val isYouth: Boolean = false,
+  val rawPncId: PNCIdentifier?,
```
markreesmoj, line 28: Surely CRO and PNC have to be stored as a String to bypass the validation and formatting?
I might be inclined to use the PNCIdentifier and CROIdentifier to do the mapping.
Is there a case for just storing this, and removing the formatted and validated version - using the presentation layer to do that work?

PR #851, opened 2025-02-24
CPR-602 store raw PNC and CRO values

samz-justice:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/identifiers/PNCIdentifierTest.kt
```diff
@@ -40,55 +40,33 @@ class PNCIdentifierTest {
...
 
-  @ParameterizedTest
-  @MethodSource("equalityPncProvider")
-  fun `should perform equality comparison using canonical form`(pncId1: String, pncId2: String, expectedResult: Boolean) {
-    // When
-    val result = PNCIdentifier.from(pncId1) == PNCIdentifier.from(pncId2)
-
-    // Then
-    assertThat(result).isEqualTo(expectedResult)
-  }
```
markreesmoj, line 51: replaced with tests to check equality of invalid and valid PNCs

H-Iwanejko (commented):
I guess we need to decide which version we are going to show to canonical version or both?

markreesmoj:
> I guess we need to decide which version we are going to show to canonical version or both?

We are going to expose the canonical version - this is write only at the moment

H-Iwanejko (approved):

PR #852, opened 2025-02-24
CPR-508-update-additional-fields-canonical

JunaedAhmedMOJ:
CPR-508-update-additional-fields-canonical added fields sex, nationality1, nationality2


H-Iwanejko (commented):
Is there plans to store the nationality2?

markreesmoj (approved):
This is fine as far as it goes - presumably we will handle writing this data to the database separately (unless it is already in there).

On another note, and this is not relevant to this PR at all, it seems a bit jarring to describe this API operation as a search. The UUID is our unique identifier so there is no searching as such - it is either there or it is not (and it will always be there unless the caller has made a mistake).

PR #853, opened 2025-02-24
CPR-584 Make matchId non nullable

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
```diff
@@ -121,7 +121,7 @@ class PersonEntity(
   val sourceSystem: SourceSystemType,
 
   @Column(name = "match_id")
-  val matchId: UUID? = null,
+  val matchId: UUID,
```
markreesmoj, line 124: All the existing rows must have a matchId for this change to work.

markreesmoj (approved):
👍 assuming all existing records in `person` have a `match_id`

PR #854, opened 2025-02-24
CPR-584 invoke person match

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/search/PersonMatchService.kt
```diff
@@ -0,0 +1,31 @@
...
+  private val personMatchClient: PersonMatchClient,
+  private val telemetryService: TelemetryService,
+  private val retryExecutor: RetryExecutor,
+) {
+
+  fun getScores(personEntity: PersonEntity): List<PersonMatchScore> = runBlocking {
+    try {
+      return@runBlocking retryExecutor.runWithRetryHTTP { personMatchClient.getPersonScores(personEntity.matchId.toString()) }
+    } catch (exception: Exception) {
+      telemetryService.trackEvent(
```
markreesmoj, line 24: Do we really need this telemetry? It will appear in AppInsights as a 500 (or whatever) and be easy to retrieve?

markreesmoj: I know it's a convention we have adopted, but I feel it is unnecessary to have a try/catch simply for this purpose

PR #858, opened 2025-02-25
CPR-602 removed identifier_raw_value from reference table

samz-justice:


JunaedAhmedMOJ (dismissed):

src/main/resources/db/migration/V83__remove_identifier_raw_value.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+-------------------------------------------------------
+
+
+ALTER TABLE reference
```
JunaedAhmedMOJ, line 5: could have IF EXISTS for consistency

samz-justice: good point.
script updated

JunaedAhmedMOJ (approved):

PR #861, opened 2025-02-25
CPR-508-map-additional-address-fields

JunaedAhmedMOJ:
CPR-508-map-additional-address-fields mapped new address fields


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/AddressEntity.kt
```diff
@@ -60,6 +87,16 @@ class AddressEntity(
         noFixedAbode = address.noFixedAbode,
         postcode = address.postcode,
         fullAddress = address.fullAddress,
+        subBuildingName = address.subBuildingName,
```
H-Iwanejko, line 90: Does if (address.postcode.isNullOrEmpty()) { still apply now?

samz-justice (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/AddressEntity.kt
JunaedAhmedMOJ: Spoke with gary about this and he said the postcode still needs the matching part but we always want to populate address.. so i guess we should remove it ?

detekt.yml
```diff
@@ -132,7 +132,7 @@ complexity:
   LongParameterList:
     active: true
     functionThreshold: 6
-    constructorThreshold: 8
+    constructorThreshold: 15
```
markreesmoj, line 135: ❓why has this changed
See line 138

markreesmoj, line 135: Maybe look at making Address a data class rather than doing this?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/AddressEntity.kt
```diff
@@ -46,22 +46,54 @@ class AddressEntity(
...
+  val buildingName: String? = null,
+
+  @Column(name = "building_number")
+  val buildingNumber: String? = null,
+
+  @Column(name = "thoroughfare_name")
+  val thoroughfareName: String? = null,
+
+  @Column(name = "dependent_locality")
+  val dependentLocality: String? = null,
```
markreesmoj, line 62: Are we quite sure about these names? They look like they come directly from the Postcode Address Format.
Are they at all meaningful to us, or will we purely be using them to populate from a PAF data source?

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/test/messages/CourtCommonPlatformMessages.kt
```diff
@@ -114,11 +118,11 @@ private fun personDefendant(commonPlatformHearingSetup: CommonPlatformHearingSet
...
-        "address2": "Cardiff",
-        "address3": "Wales",
-        "address4": "UK",
-        "address5": "Earth",
+        "address1": "${commonPlatformHearingSetup.address?.buildingName}",
+        "address2": "${commonPlatformHearingSetup.address?.buildingNumber}",
+        "address3": "${commonPlatformHearingSetup.address?.thoroughfareName}",
+        "address4": "${commonPlatformHearingSetup.address?.dependentLocality}",
+        "address5": "${commonPlatformHearingSetup.address?.postTown}",
         "postcode": "CF10 1FU"
```
markreesmoj, line 126: not part of the change I know but would be great to see this hard-coded postcode gone as well

samz-justice (approved):

PR #862, opened 2025-02-25
CPR-508-address-fields-canonical

JunaedAhmedMOJ:
CPR-508-address-fields-canonical add address fields to canonical api

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
@@ -7,6 +7,15 @@ data class CanonicalAddress(
   val startDate: String? = "",
   val endDate: String? = "",
   val postcode: String? = "",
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
```diff
@@ -7,6 +7,15 @@ data class CanonicalAddress(
   val startDate: String? = "",
   val endDate: String? = "",
   val postcode: String? = "",
+  val subBuildingName: String? = "",
+  val buildingName: String? = "",
+  val buildingNumber: String? = "",
+  val thoroughfareName: String? = "",
+  val dependentLocality: String? = "",
```
markreesmoj, line 14: As before - this is an API. We have taken this (to me) meaningless term - see also PostTown and spread it throughout our codebase, and now we are exposing it to consumers.
If this is a MoJ wide decision to do this fair enough

markreesmoj (approved):
👍

PR #885, opened 2025-03-06
CPP-634 code should now handle the case where the UUID does not retur…

samz-justice:
cases covererd:
wrong role - return 403
unauthorised - return 401
UUID is invalid -> returns 404 - BAD_REQUEST, with correct message
UUID is valid - does not exist in the db - results in empty list -> returns 400, NOT_FOUND, with correct message

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalRecord.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalRecord.kt
@@ -2,6 +2,7 @@ package uk.gov.justice.digital.hmpps.personrecord.api.model.canonical
 
 import io.swagger.v3.oas.annotations.media.ArraySchema
 import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.personrecord.api.controller.exceptions.PersonKeyNotFoundException
 import uk.gov.justice.digital.hmpps.personrecord.api.model.canonical.CanonicalIdentifierType.CRN
 import uk.gov.justice.digital.hmpps.personrecord.api.model.canonical.CanonicalIdentifierType.C_ID
 import uk.gov.justice.digital.hmpps.personrecord.api.model.canonical.CanonicalIdentifierType.DEFENDANT_ID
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalRecord.kt
```diff
@@ -45,23 +46,29 @@ data class CanonicalRecord(
...
-        ethnicity = latestPerson.ethnicity,
-        aliases = CanonicalAlias.fromPseudonymEntityList(latestPerson.pseudonyms),
-        addresses = CanonicalAddress.fromAddressEntityList(latestPerson.addresses),
-        identifiers = getCanonicalIdentifiers(personKey.personEntities) + CanonicalIdentifier.fromReferenceEntityList(latestPerson.references),
-        nationalities = CanonicalNationality.from(latestPerson),
+      val personList = personKey.personEntities.sortedByDescending { it.lastModified }
 
-      )
+      when (personList.isNullOrEmpty()) {
+        true -> throw personKey.personId?.let { PersonKeyNotFoundException(it) }!!
```
grnewbould, line 52: General Comment.  Is there an alternative to handle this rather than exceptions.  Exceptions have an expense on the jvm as they collect stats, stack traces etc ? Do we use exceptions for business errors elsewhere ?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalRecord.kt
```diff
@@ -45,23 +46,29 @@ data class CanonicalRecord(
...
-        religion = latestPerson.religion,
-        ethnicity = latestPerson.ethnicity,
-        aliases = CanonicalAlias.fromPseudonymEntityList(latestPerson.pseudonyms),
-        addresses = CanonicalAddress.fromAddressEntityList(latestPerson.addresses),
-        identifiers = getCanonicalIdentifiers(personKey.personEntities) + CanonicalIdentifier.fromReferenceEntityList(latestPerson.references),
-        nationalities = CanonicalNationality.from(latestPerson),
+      val personList = personKey.personEntities.sortedByDescending { it.lastModified }
 
-      )
+      when (personList.isNullOrEmpty()) {
```
markreesmoj, line 51: I think there might be a better place for this logic

https://github.com/ministryofjustice/hmpps-person-record/blob/a603699a9e2619c1865b401a23eef46fe39fc0a5/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/SearchController.kt#L118

we are already checking if the entity exists here (incorrectly BTW - please also fix the check!) so we could just go one step further and check the list of people is not empty too?

Something like this should do the trick?

`  private fun buildCanonicalRecord(personKeyEntity: PersonKeyEntity?, uuid: UUID): CanonicalRecord = when {
    personKeyEntity?.personEntities?.isNotEmpty() == true -> CanonicalRecord.from(personKeyEntity)
    else -> throw PersonKeyNotFoundException(uuid)
  }`

markreesmoj (commented):
Tests looking good, it doesn't seem quite right for the companion object of a `model` to be throwing an exception from the `controller` package

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/controller/SearchIntTest.kt
```diff
@@ -569,6 +569,44 @@ class SearchIntTest : WebTestBase() {
...
+  @Test
+  fun `should return bad request with userMessage to show that the UUID is not found`() {
+    val randomUUId = "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
+    val expectedErrorMessage = "Not found: $randomUUId"
+    webTestClient.get()
+      .uri(searchForPerson(randomUUId))
+      .authorised(listOf(SEARCH_API_READ_ONLY))
+      .exchange()
+      .expectStatus()
+      .isNotFound
```
markreesmoj, line 581: and this says not found (404)

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/controller/SearchIntTest.kt
```diff
@@ -569,6 +569,44 @@ class SearchIntTest : WebTestBase() {
       .isBadRequest
   }
 
+  @Test
+  fun `should return bad request with userMessage to show that the UUID is not found`() {
```
markreesmoj, line 573: this says bad request (400)

markreesmoj: test just needs renaming, the bad request test is above

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalRecord.kt
markreesmoj: Incidentally I have tested the API in dev and this functionality already works as expected

markreesmoj (approved):
👍

PR #894, opened 2025-03-11
CPR-644-resolve-404-person-match

JunaedAhmedMOJ:
call retry for person match call


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/RetryExecutor.kt
```diff
@@ -81,4 +85,5 @@ class RetryExecutor(@Value("\${retry.delay}") val delayMillis: Long) {
   )
 
   private val httpRetryExceptions = listOf(feign.RetryableException::class, FeignException.InternalServerError::class, FeignException.ServiceUnavailable::class, FeignException.BadGateway::class)
+  private val httpRetryException404 = listOf(FeignException.NotFound::class)
```
markreesmoj, line 88: Should probably be all the above retry exceptions plus this one, rather than either/or

H-Iwanejko: We get 404s from delius on some events, which we don't want to retry on / increase processing time.

PR #900, opened 2025-03-12
CPR-629 added crn and prison number to PersonMatchRecord data class

samz-justice:


JunaedAhmedMOJ:
Could we include an integration test to prove crn is posted to person match?

JunaedAhmedMOJ:
> Could we include an integration test to prove crn is posted to person match?

nevermind looks like we do already have one :)

H-Iwanejko (approved):

PR #924, opened 2025-03-21
CPR-603 set bucket to enable extended client

markreesmoj:


helm_deploy/hmpps-person-record/values.yaml
```diff
@@ -95,8 +95,7 @@ generic-service:
     sqs-cpr-recluster-events-dlq-secret:
       HMPPS_SQS_QUEUES_CPRRECLUSTEREVENTSQUEUE_DLQ_NAME: "sqs_queue_name"
     large-court-cases-s3-bucket:
-      AWS_COURT_MESSAGE_BUCKET_NAME: "bucket_name"
```
markreesmoj, line 98: This is what we missed yesterday - setting this property configures the SQS extended client and without it we will get the error we saw yesterday:

```2025-03-20 16:44:15.998  INFO 1 --- [ntContainer#0-1] u.g.j.h.sqs.HmppsQueueFactory$Companion  : Setting visibility of messageId aac140e4-cefa-422c-9170-6f900615fac3 to 0 (to initiate faster retry) after receiving exception com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `uk.gov.justice.digital.hmpps.personrecord.client.model.court.event.CommonPlatformHearingEvent` from Array value (token `JsonToken.START_ARRAY`)
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 1] | trace_id=b1cf77d21f28dc59632d6314e557b817, trace_flags=01, span_id=4c3929be90090adc 
2025-03-20 16:44:16.027 ERROR 1 --- [ntContainer#0-1] .s.AbstractMessageProcessingPipelineSink : Error processing message aac140e4-cefa-422c-9170-6f900615fac3. |  

java.util.concurrent.CompletionException: io.awspring.cloud.sqs.listener.AsyncAdapterBlockingExecutionFailedException: Error executing action in BlockingErrorHandlerAdapter```

samz-justice (approved):

PR #930, opened 2025-03-24
CPR-615 Add changed record to recluster queue message

H-Iwanejko:
* Add change record Id do recluster queue to highlight changed record to trigger new recluster process from.
* Let kotlin do UUID encode / decode

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/queue/QueueService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/queue/QueueService.kt
@@ -8,6 +8,7 @@ import uk.gov.justice.digital.hmpps.personrecord.client.model.sqs.NOTIFICATION
 import uk.gov.justice.digital.hmpps.personrecord.client.model.sqs.messages.Recluster
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.PersonEntity
 import uk.gov.justice.digital.hmpps.personrecord.service.type.RECLUSTER_EVENT
+import uk.gov.justice.hmpps.sqs.HmppsQueue
 import uk.gov.justice.hmpps.sqs.HmppsQueueService
 import uk.gov.justice.hmpps.sqs.MissingQueueException
 import java.util.UUID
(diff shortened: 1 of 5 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/cpr/ReclusterEventListenerIntTest.kt
```diff
@@ -91,9 +91,10 @@ class ReclusterEventListenerIntTest : MessagingMultiNodeTestBase() {
 
   @Test
   fun `should recluster when single record matches to one other cluster`() {
+    telemetryRepository.deleteAllInBatch()
```
H-Iwanejko, line 94: Need this to solve flakyness of candidate search, will get remove and solved when used new person-match results

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/queue/QueueService.kt
```diff
@@ -18,10 +19,13 @@ class QueueService(
   private val hmppsQueueService: HmppsQueueService,
 ) {
 
-  fun publishReclusterMessageToQueue(uuid: UUID) {
+  fun publishReclusterMessageToQueue(personEntity: PersonEntity) {
     val queue = findByQueueIdOrThrow(Queues.RECLUSTER_EVENTS_QUEUE_ID)
```
markreesmoj, line 23: I know this is not going to be around very long but can we do this call when we initialise the class instead? It is not going to change

H-Iwanejko: Not sure what you mean? All this does is push a message the recluster queue

markreesmoj: https://github.com/ministryofjustice/court-hearing-event-receiver/blob/fc22f02e5bfe8d66dc553bb85fd635d33fda7125/src/main/kotlin/uk/gov/justice/digital/hmpps/courthearingeventreceiver/service/MessageNotifier.kt#L29

I mean this sort of thing - instead of calling findByQueueIdOrThrow multiple times, just do it once

markreesmoj (approved):
👍

PR #940, opened 2025-03-26
CPR-652 resolve with retry

JunaedAhmedMOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/CreateUpdateService.kt
```diff
@@ -25,7 +26,7 @@ class CreateUpdateService(
   private val eventLoggingService: EventLoggingService,
 ) {
 
-  @Transactional
+  @Transactional(isolation = Isolation.SERIALIZABLE)
```
H-Iwanejko, line 29: There is a sub `Transactional` defined for reclustering with this `Transaction`. Might be good to make it handle them subsequently rather than nested

markreesmoj: I've made it part of the same transaction - needed to modify the tests to use messaging in order to achieve this

markreesmoj:
Removed all retries apart from create/update processing
Used Spring retries instead of the custom code

H-Iwanejko (approved):
Lets give it ago 👍

markreesmoj (approved):
👍

PR #942, opened 2025-03-27
CPR-615 new recluster no change + needs attention in composition

H-Iwanejko:
* Start calling person-match for recluster results


markreesmoj (approved):
Good start

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceIntTest.kt
```diff
@@ -2,502 +2,212 @@ package uk.gov.justice.digital.hmpps.personrecord.service
```
markreesmoj, line 2: As reclustering now happens solely as a result of update processing, I think it would be sensible to have at least one test which proves the whole process - consume an update message and verify that reclustering happens as a result.

PR #944, opened 2025-03-27
CPR-652 simplify code

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/CreateUpdateService.kt
```diff
@@ -26,17 +36,31 @@ class CreateUpdateService(
...
+    retryFor = [
+      ObjectOptimisticLockingFailureException::class,
+      CannotAcquireLockException::class,
+      JpaSystemException::class,
+      JpaObjectRetrievalFailureException::class,
+      DataIntegrityViolationException::class,
+      ConstraintViolationException::class,
+      StaleObjectStateException::class,
+    ],
+  )
```
markreesmoj, line 50: See https://stackoverflow.com/questions/49678581/spring-retry-with-transactional - this seems to indicate that @Retryable will happen around @Transactional

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/CreateUpdateService.kt
```diff
@@ -26,17 +30,25 @@ class CreateUpdateService(
   private val eventLoggingService: EventLoggingService,
 ) {
 
+  @Retryable(
+    backoff = Backoff(random = true, delay = 1000, maxDelay = 2000, multiplier = 1.5),
+    retryFor = [
+      OptimisticLockException::class,
```
markreesmoj, line 36: This one is definitely happening in dev :-)

JunaedAhmedMOJ (approved):

PR #948, opened 2025-03-28
CPR-616 Remove old candidate search and person match score results

H-Iwanejko:
* Remove code to call personMatchScore
* Remove old candidate search
* Remove associated object code
* Remove personMatchScore from health check


markreesmoj (approved):
✅

PR #950, opened 2025-03-28
CPR-79 Increase db connection pool size

H-Iwanejko:
* Should resolve the small amount of errors related to: `HikariPool-1 - Connection is not available, request timed out after 30000ms (total=10, active=10, idle=0, waiting=3)`


markreesmoj (approved):
👍

PR #967, opened 2025-04-03
CPR-616 Send as a list on in a built json

H-Iwanejko:


samz-justice (approved):

markreesmoj (commented):
OK but surely there should be a change to some test stubbing to match this? Or is it not set up to match the post body accurately?

PR #991, opened 2025-04-10
CPR-489 inject UUID

JunaedAhmedMOJ:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/court/libra/LibraCourtEventListenerIntTest.kt
```diff
@@ -207,6 +207,7 @@ class LibraCourtEventListenerIntTest : MessagingMultiNodeTestBase() {
     val libraMessage: String = sqsMessage?.message!!
 
     assertThat(libraMessage.contains(cId)).isEqualTo(true)
+    assertThat(libraMessage.contains("cprUUID")).isEqualTo(true)
```
markreesmoj, line 210: assert the value of the UUID

markreesmoj (commented):
defendantId does not exist on LIBRA cases, use CId

PR #1009, opened 2025-04-22
Rename and reuse

markreesmoj:
What do we all think of this?

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
@@ -227,9 +227,9 @@ class PersonEntity(
 
     fun List<ReferenceEntity>.getType(type: IdentifierType): List<ReferenceEntity> = this.filter { it.identifierType == type }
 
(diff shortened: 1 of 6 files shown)


H-Iwanejko (approved):
Like this, makes event logging easier ✅

markreesmoj:
What I think:

I like the renaming from `shouldCreateOrUpdate:shouldCreate` to `exists:yes`
I am less sure about the side effect of setting the key on the personEntity in the personKeyService 
I also do not really like returning the personEntity where it is in fact the modified entity which was originally passed in (this is a general pattern in our code)

I feel a possible cure might be to do away with `PersonKeyService` because
1. It is only called from PersonService.
2. it shares two out of its three dependencies with it 
3. it would reduce the need to return a modified personEntity

Would love to hear other views on this

JunaedAhmedMOJ (approved):

H-Iwanejko:
>  I also do not really like returning the personEntity where it is in fact the modified entity which was originally passed in (this is a general pattern in our code)

The issue is if we dont do this the changes dont get propagated down the chain (unfortunate) annoying and need to investigate how to solve this. 

But in generally agree with moving this into the person service as its a very basic service regardless.

PR #1012, opened 2025-04-23
CPR-610 Add UUID merged event log

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/MergeService.kt
```diff
@@ -108,10 +110,13 @@ class MergeService(
...
     sourcePersonEntity?.let {
+      it.hasPersonKey(
+        yes = { personKeyEntity -> personKeyRepository.save(personKeyEntity) },
+        no = { personEntity -> personRepository.save(personEntity) },
+      )
       publisher.publishEvent(RecordEventLog(CPRLogEvents.CPR_RECORD_MERGED, it))
-      personRepository.save(it)
     }
-    personRepository.save(targetPersonEntity)
+    targetPersonEntity.personKey?.let { personKeyRepository.save(it) }
```
H-Iwanejko, line 119: Save from the personKey rather than person entity as it propagates changes anyway. Also requires less saving

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/MergeService.kt
```diff
@@ -58,8 +60,8 @@ class MergeService(
 
   private fun handleSourceUuidWithSingleRecord(mergeEvent: MergeEvent, sourcePersonEntity: PersonEntity, targetPersonEntity: PersonEntity) {
     mergeRecord(mergeEvent, sourcePersonEntity, targetPersonEntity) { sourcePerson, targetPerson ->
-      linkSourceUuidToTargetAndMarkAsMerged(sourcePerson!!, targetPerson)
-      updateAndLinkRecords(mergeEvent, sourcePerson, targetPerson)
+      updateAndLinkRecords(mergeEvent, sourcePerson!!, targetPerson)
+      linkSourceUuidToTargetAndMarkAsMerged(sourcePerson, targetPerson)
```
H-Iwanejko, line 64: Swapped around so event log contains the mergedTo link

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
```diff
@@ -144,12 +149,15 @@ class PersonEntity(
...
+  fun mergedTo(personEntity: PersonEntity) {
+    this.mergedTo = personEntity.id
+  }
+
   fun removeMergedLink() {
     this.mergedTo = null
   }
 
   fun removePersonKeyLink() {
-    this.personKey?.personEntities?.remove(this)
```
H-Iwanejko, line 152: Not required when saving from personKey as it propagates

markreesmoj: https://github.com/ministryofjustice/hmpps-person-record/blob/0007fcb3692bb81f07e00a70cfe120c5c40f55eb/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/UnmergeService.kt#L94

Have you seen this? This is calling `removePersonKeyLink` but saving it via the PersonRepository

Will this still work as expected after the above change?

markreesmoj (commented):
💬

PR #1013, opened 2025-04-23
CPR-668 rename col sex to sex_code on the person table

samz-justice:
this PR is to only rename the sex col on the person table to sex_code


JunaedAhmedMOJ (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
```diff
@@ -104,7 +104,7 @@ class PersonEntity(
   @Column(name = "date_of_birth")
   var dateOfBirth: LocalDate? = null,
 
-  @Column
+  @Column(name = "sex_code")
   val sex: String? = null,
```
H-Iwanejko, line 108: Does this want to be `sexCode` now?

PR #1014, opened 2025-04-23
CPR-654-circular-merges

JunaedAhmedMOJ:
add merged to check constraint


samz-justice (approved):

markreesmoj:
Only 2 records in production affected by this

PR #1015, opened 2025-04-23
CPR-668 added sex_description to person table

samz-justice:


markreesmoj:
Do we need to store this? I thought it would be the description in an enumeration containing sex_code

samz-justice:
closing this as it not required after a 3 amigos session with the BA

PR #1018, opened 2025-04-23
CPR-610 Add new merge service

H-Iwanejko:
* add new merge service to incrementally replace the old one
* No telemetry / event log introduced (yet)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/NewMergeService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/NewMergeService.kt
@@ -15,8 +15,8 @@ class NewMergeService(
 
   @Transactional
   fun processMerge(from: PersonEntity?, to: PersonEntity) {
-    if (fromRecordIsAloneOnSeparateCluster(from, to)) {
-      markClusterAsMerged(from, to)
+    when {
+      fromClusterHasOneRecord(from) -> markClusterAsMerged(from, to)
     }
     merge(from, to)
   }
@@ -37,9 +37,5 @@ class NewMergeService(
     }
   }
 
-  private fun fromRecordIsAloneOnSeparateCluster(from: PersonEntity?, to: PersonEntity): Boolean = fromClusterHasOneRecord(from) and isDifferentUuid(from, to)
-
   private fun fromClusterHasOneRecord(from: PersonEntity?): Boolean = (from?.personKey?.personEntities?.size ?: 0) == 1
-
-  private fun isDifferentUuid(from: PersonEntity?, to: PersonEntity): Boolean = from?.personKey?.id != to.personKey?.id
 }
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/NewMergeService.kt
```diff
@@ -0,0 +1,45 @@
...
+  private fun merge(from: PersonEntity?, to: PersonEntity) {
+    from?.let {
+      it.removePersonKeyLink()
+      it.mergeTo(to)
+      personRepository.save(it)
+      deletionService.deletePersonFromPersonMatch(it)
+    }
+  }
+
+  private fun fromRecordIsAloneOnSeparateCluster(from: PersonEntity?, to: PersonEntity): Boolean = fromClusterHasOneRecord(from) and isDifferentUuid(from, to)
```
markreesmoj, line 40: If the from cluster only has one record, surely the to cluster cannot by definition have the same UUID

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/NewMergeServiceIntTest.kt
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @Test
+  fun `should merge records on different UUIDs with single records`() {
+    val from = createPersonWithNewKey(createRandomProbationPersonDetails())
+    val to = createPersonWithNewKey(createRandomProbationPersonDetails())
+
+    newMergeService.processMerge(from, to)
+
+    from.personKey?.assertClusterStatus(UUIDStatusType.MERGED)
```
markreesmoj, line 27: interesting that we do not have to read the records from the database again - this works because the `processMerge` method is transactional I assume.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/NewMergeServiceIntTest.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    cluster.assertClusterStatus(UUIDStatusType.ACTIVE)
+    cluster.assertClusterIsOfSize(1)
+
+    from.assertMergedTo(to)
+    from.assertNotLinkedToCluster()
+  }
+
+  @Test
+  fun `should merge records on same different UUIDs with multiple records`() {
```
markreesmoj, line 57: same different?
It's hard to express what this test is proving though!
should move record from cluster with multiple records without merging the whole cluster?

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/NewMergeServiceIntTest.kt
H-Iwanejko: It grabs the latests changes in the helper method 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/NewMergeServiceIntTest.kt
H-Iwanejko: Updated 👍

samz-justice (approved):

PR #1020, opened 2025-04-24
CPR-662 using common platform message to populate the sexcode field

samz-justice:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/types/SexCodeTest.kt
```diff
@@ -0,0 +1,102 @@
...
+
+      assertEquals(SexCode.from(personDetails), null)
+    }
+  }
+
+  @Nested
+  inner class FromLibraCourtEvent {
+
+    @Test
+    fun `should map from M defendantSex to M sexCode`() {
```
H-Iwanejko, line 68: These tests can all be put under one parameterized test

markreesmoj (approved):
👍

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/types/SexCodeTest.kt
samz-justice: updated

markreesmoj (approved):
👍

PR #1034, opened 2025-04-25
CPR-647 add deletion event log

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/DeletionService.kt
```diff
@@ -1,63 +1,48 @@
...
-  }
+  @Transactional
+  fun processDelete(event: String?, personCallback: () -> PersonEntity?) = fetchRecordAndDelete(event, personCallback)
 
-  fun deletePersonFromPersonMatch(personEntity: PersonEntity) = runCatching {
-    runBlocking {
-      retryExecutor.runWithRetryHTTP { personMatchClient.deletePerson(PersonMatchIdentifier.from(personEntity)) }
-    }
-  }
+  private fun fetchRecordAndDelete(event: String?, personCallback: () -> PersonEntity?) = personCallback()?.let { handleDeletion(event, it) }
```
H-Iwanejko, line 25: To make a transactional had to separate into different functions to allow for recursion within a transaction. Without opening multiple sub transactions

samz-justice (approved):

PR #1040, opened 2025-04-29
CPR-647 Add cluster composition to events log

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/listeners/EventLogListener.kt
```diff
@@ -16,6 +16,6 @@ class EventLogListener(
   @EventListener
   @TransactionalEventListener
   fun onEventLogEvent(eventLog: RecordEventLog) {
-    eventLogService.logEvent(eventLog.personEntity, eventLog.eventType, eventLog.personKeyEntity)
+    eventLogService.logEvent(eventLog.personEntity, eventLog.eventType, eventLog.personKeyEntity, eventLog.clusterComposition)
```
samz-justice, line 19: it might be better to pass eventLog and let the logEvent function grab what it needs internally

markreesmoj (approved):
👍

PR #1046, opened 2025-04-30
CPR-610 Plumb new unmerge service into probation

H-Iwanejko:
* Subsequent PR coming to rename `reactivated` and `unmerged` -> `from` and `to`


markreesmoj (approved):
👍

PR #1049, opened 2025-04-30
CPR-610 Fix null as string cluster composition

H-Iwanejko:
* Fixes bug that stores cluster composition as "null" to just null


markreesmoj (approved):
👍

PR #1056, opened 2025-05-07
CPR-669 Send source system ID to person match

H-Iwanejko:
* Send source system id
* Use and store source system id on person match
* Then will remove the crn + prison number from transaction

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/match/PersonMatchRecord.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/match/PersonMatchRecord.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.personrecord.client.model.match
 
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.PersonEntity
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/match/PersonMatchRecord.kt
```diff
@@ -20,6 +21,7 @@ data class PersonMatchRecord(
   val sentenceDates: List<String> = listOf(),
   val crn: String? = "",
   val prisonNumber: String? = "",
+  val sourceSystemId: String? = "",
```
samz-justice, line 24: Q: isn't a mandatory to have sourceSystemId to be available all the time, hence the initial value so an empty string is not right?

H-Iwanejko: The `extractSourceSystemId` returns a null just a default value, will never be null in this scenario

H-Iwanejko: Stops the detekt from firing `TooComplex` when just specify `?: ""`

samz-justice (approved):
isn't true that sourceSystemId

PR #1057, opened 2025-05-07
CPR-689 only store person with firstName

samz-justice:


H-Iwanejko (approved):
✅ If the minimun data present is agreed upon + confirmed by BA's that we can do this

markreesmoj (approved):
👍

markreesmoj (approved):
👍

PR #1079, opened 2025-05-16
CPR-710 reclustering needs attention clusters which are now valid

samz-justice:
…ctive cluster


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/recluster/ReclusterService.kt
```diff
@@ -29,20 +30,29 @@ class ReclusterService(
 ) {
 
   fun recluster(cluster: PersonKeyEntity, changedRecord: PersonEntity) {
-    when {
-      clusterNeedsAttention(cluster) -> publisher.publishEvent(RecordClusterTelemetry(TelemetryEventType.CPR_RECLUSTER_UUID_MARKED_NEEDS_ATTENTION, cluster))
-      else -> handleRecluster(cluster, changedRecord)
+    if (clusterNeedsAttentionAndIsInvalid(cluster, changedRecord)) {
```
JunaedAhmedMOJ, line 33: perhaps seperate the check and the action rather than trying to do both here

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/recluster/ReclusterService.kt
```diff
@@ -126,6 +162,19 @@ class ReclusterService(
     personKeyRepository.save(clusterDetails.cluster)
   }
 
+  private fun clusterIsActive(personKeyEntity: PersonKeyEntity, changedRecord: PersonEntity) {
```
JunaedAhmedMOJ, line 165: Might be better to change method to setClusterIsActive?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/recluster/ReclusterService.kt
```diff
@@ -29,20 +31,29 @@ class ReclusterService(
...
+    val matchesToChangeRecord: List<PersonMatchResult> =
+      personMatchService.findHighestConfidencePersonRecordsByProbabilityDesc(changedRecord)
     val clusterDetails = ClusterDetails(cluster, changedRecord, matchesToChangeRecord)
+
     when {
       clusterDetails.relationship.isDifferent() -> handleDiscrepancyOfMatchesToExistingRecords(clusterDetails)
     }
   }
 
+  private fun clusterNeedsAttentionAndIsInvalid(cluster: PersonKeyEntity) = cluster.isNeedsAttention() && !personMatchService.examineIsClusterValid(cluster).isClusterValid
```
H-Iwanejko, line 55: This is now calling `isClusterValid` everytime regardless if its needs attention or not, as it tries to resolve the conditional. Need to change this

markreesmoj: according to https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-boolean/and.html it will evaluate `isNeedsAttention` first and if false, will not evaluate the other side of the `&&`

markreesmoj (commented):
💬

PR #1090, opened 2025-05-21
CPR-663 Further listener refactor

H-Iwanejko:
* Once probation is using all domain events we can utilise these helper funcs better

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/queue/SQSListenerService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/queue/SQSListenerService.kt
@@ -27,12 +27,13 @@ class SQSListenerService(
   companion object {
     private val log = LoggerFactory.getLogger(this::class.java)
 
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/queue/SQSListenerService.kt
```diff
@@ -2,20 +2,45 @@ package uk.gov.justice.digital.hmpps.personrecord.service.queue
...
+      return this
+    }
+
+    fun DomainEvent.whenEvent(eventType: String, action: (domainEvent: DomainEvent) -> Unit): DomainEvent {
+      when {
+        this.eventType == eventType -> action(this)
+      }
+      return this
+    }
+  }
```
markreesmoj, line 45: I'm sure you thought about this, just curious to know why these two are here rather than in the DomainEvent?

markreesmoj (approved):
Looking very good

PR #1092, opened 2025-05-22
CPR-663 Split court into seperate processors

H-Iwanejko:


markreesmoj (approved):
A step in the right direction

PR #1098, opened 2025-05-23
CPR-622 added match report endpoint to the cron job

samz-justice:


markreesmoj (approved):
Not sure this will work - might have to either repeat the 
`- /bin/sh`
` -c`

or use `&&` and put both `curl` commands on the same line if it doesn't

PR #1099, opened 2025-05-23
CPR-393 remove custom oauth config

markreesmoj:
a big job


markreesmoj:
Have a look at this on Tuesday
https://github.com/ministryofjustice/hmpps-kotlin-lib/blob/main/readme-contents/TestHelpers.md

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/IntegrationTestBase.kt
```diff
@@ -102,7 +123,7 @@ class IntegrationTestBase {
             wiremock.verify(getRequestedFor(urlMatching(it.request.urlPathPattern)))
           }
         }
-        POST -> wiremock.verify(postRequestedFor(urlEqualTo(it.request.url)))
+        POST -> Unit // wiremock.verify(postRequestedFor(urlEqualTo(it.request.url)))
```
markreesmoj, line 126: would like to work on this - the current approach is not ideal. It seems like the stubbing sometimes happens when it is not needed. Unsure how this can be as it is only called when another stub is being set up, which was already being verified

PR #1106, opened 2025-05-30
CPR-745-rename-uuid

JunaedAhmedMOJ:
rename uuid column in event_log to person_uuid


markreesmoj:
We are doing this because it is annoying to have to write double quotes around uuid in dBeaver - it is a reserved word I think

H-Iwanejko (approved):

PR #1116, opened 2025-06-03
CPR-640 use retryable annotation for HTTP errors from prisons

markreesmoj:


JunaedAhmedMOJ (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/CreateUpdateService.kt
```diff
@@ -28,6 +32,10 @@ class CreateUpdateService(
       OptimisticLockException::class,
       DataIntegrityViolationException::class,
       CannotAcquireLockException::class,
+      InternalServerError::class,
+      BadGateway::class,
+      ServiceUnavailable::class,
+      WebClientRequestException::class,
```
H-Iwanejko, line 38: Is it worth split these errors into specific lists and merging? 
i.e. DBRetry = x,y,z.   HttpRetry = a,b,c      then just DBRetry + HttpRetry.

H-Iwanejko (approved):
Worth a try

PR #1121, opened 2025-06-04
CPR-717 Remove currently managed

H-Iwanejko:


markreesmoj (changes requested):
Please do the DB column removal in a separate, later PR
Doing it this way will result in errors as the pod with the existing code on will still be processing prison messages and will error once the new deployment has come up and deleted the column
We are a live service now and do not want any unnecessary errors

H-Iwanejko:
Removed Migration 👍

PR #1126, opened 2025-06-04
CPR-640 no need to use retryexecutor for person match calls

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/search/PersonMatchService.kt
```diff
@@ -57,11 +55,14 @@ class PersonMatchService(
     )
   }
 
-  fun saveToPersonMatch(personEntity: PersonEntity): ResponseEntity<Void> = runBlocking { retryExecutor.runWithRetryHTTP { personMatchClient.postPerson(PersonMatchRecord.from(personEntity)) } }
+  fun saveToPersonMatch(personEntity: PersonEntity): ResponseEntity<Void>? = personMatchClient.postPerson(PersonMatchRecord.from(personEntity))
 
-  fun deleteFromPersonMatch(personEntity: PersonEntity) = runBlocking { runCatching { retryExecutor.runWithRetryHTTP { personMatchClient.deletePerson(PersonMatchIdentifier.from(personEntity)) } } }
+  fun deleteFromPersonMatch(personEntity: PersonEntity): ResponseEntity<Void>? = personMatchClient.deletePerson(PersonMatchIdentifier.from(personEntity))
```
H-Iwanejko, line 60: This should not throw errors on a 404 when deleting, for what ever reason that record is not there already just ignore it i think, hence the test failing

PR #1139, opened 2025-06-09
CPR-751 Log when there are missing keys for common platform messages

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/court/CommonPlatformEventProcessor.kt
```diff
@@ -36,6 +41,26 @@ class CommonPlatformEventProcessor(
       else -> sqsMessage.message
     }
 
+    val commonPlatformHearingNode = objectMapper.readTree(commonPlatformHearing)
+
+    commonPlatformHearingNode.path("hearing").path("prosecutionCases")
+      .flatMap { it.path("defendants") }
+      .filter {
+        it.path("isYouth").isMissingNode and
```
H-Iwanejko, line 49: Should this be an `and` or an `or`?

PR #1150, opened 2025-06-12
CPR-640 set message visibility on listeners

markreesmoj:
https://github.com/awspring/spring-cloud-aws/blob/0c1cec285bdceb11d1e5cfa2c9d8b292cee7c859/spring-cloud-aws-sqs/src/main/java/io/awspring/cloud/sqs/annotation/SqsListener.java#L133-L138

I think this will have the same effect, and it uses the framework code rather than our own.


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/court/CourtEventListener.kt
```diff
@@ -18,7 +18,7 @@ class CourtEventListener(
   private val sqsListenerService: SQSListenerService,
 ) {
 
-  @SqsListener(Queues.COURT_CASES_QUEUE_ID, factory = "hmppsQueueContainerFactoryProxy")
+  @SqsListener(Queues.COURT_CASES_QUEUE_ID, factory = "hmppsQueueContainerFactoryProxy", messageVisibilitySeconds = "30")
```
H-Iwanejko, line 21: the `messageVisibilitySeconds` means the SQS hides the message temporarily so no other consumer will process it. It doesn't act as a timeout for the process itself, i think. 

If the process takes in this example longer than 30 secs here. that message will be visible on the queue again and repicked up. But the message will still be processing the old one causing duplication of processing of messages.

Our sqs `visibility_timeout_seconds` is 120 seconds. 

The retry executor: `Rule of thumb: Message timeout <= visibility_timeout_seconds / maxReceiveCount` is a defensive solution to the scenario above.

I don't think we achieve anything by doing this.

markreesmoj:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html
Suggests that the timeout is 30 seconds anyway so this does not achieve anything

PR #1152, opened 2025-06-12
CRP-640 supressing wanrning from SqlExceptionHelper

samz-justice:


samz-justice:
this is meant to suppress the Warnings

markreesmoj:
> i know we get a lot of:`ERROR: could not serialize access due to concurrent update `
> 
> But would this inadvertly suppress future SQL errors we actually care about?

this suppresses warnings. We will still see the errors

H-Iwanejko:
> > i know we get a lot of:`ERROR: could not serialize access due to concurrent update `
> > But would this inadvertly suppress future SQL errors we actually care about?
> 
> this suppresses warnings. We will still see the errors

I dont know much about the log stuff but its set to `level="ERROR"`

PR #1175, opened 2025-06-19
CPR-699 Add summary to admin recluster

H-Iwanejko:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/controller/admin/AdminApiIntTest.kt
```diff
@@ -85,8 +88,11 @@ class AdminApiIntTest : WebTestBase() {
         .isOk
 
       checkTelemetry(
-        TelemetryEventType.CPR_ADMIN_RECLUSTER_TRIGGERED,
-        mapOf("UUID" to person.personKey?.personUUID.toString()),
+        TelemetryEventType.CPR_ADMIN_RECLUSTER_SUMMARY,
+        mapOf(
+          "ACTIVE -> ACTIVE" to "1",
```
markreesmoj, line 93: This is on a cluster level, right?
So where we have a NEEDS_ATTENTION cluster with 3 records, and the first update causes it to become ACTIVE, this would report

> "NEEDS_ATTENTION -> ACTIVE" to "1",
> "ACTIVE -> ACTIVE" to "2",

Is that correct? If so, that does not really reflect what happened here.

Maybe we should get together and spend a bit of time defining what we want to report here?

H-Iwanejko: Sounds good, this is a tricky one 🤔

PR #1212, opened 2025-07-07
reducing the paramaters

samz-justice:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/recluster/ReclusterService.kt
```diff
@@ -149,22 +149,23 @@ class ReclusterService(
...
       ),
     )
-    settingClusterToNeedsAttention(clusterDetails)
+    settingClusterToNeedsAttention(clusterDetails.cluster)
   }
 
-  private fun settingClusterToNeedsAttention(clusterDetails: ClusterDetails) {
-    clusterDetails.cluster.status = NEEDS_ATTENTION
-    personKeyRepository.save(clusterDetails.cluster)
+  private fun settingClusterToNeedsAttention(cluster: PersonKeyEntity) {
```
markreesmoj, line 155: I like this

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/recluster/ReclusterService.kt
```diff
@@ -149,22 +149,23 @@ class ReclusterService(
...
+    cluster.status = NEEDS_ATTENTION
+    personKeyRepository.save(cluster)
   }
 
-  private fun settingNeedsAttentionClusterToActive(personKeyEntity: PersonKeyEntity, changedRecord: PersonEntity) {
-    personKeyEntity.status = ACTIVE
-    personKeyRepository.save(personKeyEntity)
+  private fun settingNeedsAttentionClusterToActive(changedRecord: PersonEntity) {
```
markreesmoj, line 160: I am a bit less sure about this. We have both `changedRecord` and `cluster` available when calling this method, why not pass them both in?
Especially since the important change is being made on the cluster rather than the record, which is only in here for the event logging

PR #1218, opened 2025-07-08
CPR-792 failing test

markreesmoj:


samz-justice (approved):
good to have some real e2e tests

PR #1220, opened 2025-07-08
CPR-792 added unmerge controller

markreesmoj:


src/main/resources/application-local.yml
```diff
@@ -23,37 +23,42 @@ hmpps.sqs:
       queueName: cpr_court_cases_queue.fifo
       subscribeTopicId: courtcasestopic
       dlqName: cpr_court_cases_queue_dlq.fifo
-      dlqMaxReceiveCount: 1
+      dlqMaxReceiveCount: 2
```
markreesmoj, line 26: This is to match application-test so that the same localstack container can be used with both test and local running

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/court/commonplatform/CreateCommonPlatformPersonE2ETest.kt
```diff
@@ -27,7 +27,7 @@ class CreateCommonPlatformPersonE2ETest : MessagingTestBase() {
       commonPlatformHearing(listOf(CommonPlatformHearingSetup(pnc = pnc, firstName = firstName, lastName = lastName, cro = cro, defendantId = defendantId))),
     )
 
-    val updatedPersonEntity = awaitNotNullPerson(timeout = 5, function = { personRepository.findByDefendantId(defendantId) })
+    val updatedPersonEntity = awaitNotNullPerson(timeout = 7, function = { personRepository.findByDefendantId(defendantId) })
```
markreesmoj, line 30: this was failing occasionally with a timeout of 5 seconds

samz-justice (approved):

PR #1227, opened 2025-07-16
CPR-797 preserve CRO and PNC on courts messages which do not have them

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/court/CommonPlatformEventProcessor.kt
```diff
@@ -92,4 +100,19 @@ class CommonPlatformEventProcessor(
     }
     return messageParser.jsonString()
   }
+
+  private fun getIdentifiersFromDefendant(defendant: Defendant): Defendant {
+    if (defendant.isPncMissing) {
+      val getDefendant = defendant.id?.let { personRepository.findByDefendantId(it) }
```
JunaedAhmedMOJ, line 106: gets used twice, use once outside of if

PR #1232, opened 2025-07-18
Cpr 766 e2e

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/CreateUpdateService.kt
```diff
@@ -43,12 +41,6 @@ class CreateUpdateService(
   private fun update(person: Person, existingPersonEntity: PersonEntity): PersonEntity {
     val personUpdated = personService.updatePersonEntity(person, existingPersonEntity)
     publisher.publishEvent(personUpdated)
```
markreesmoj, line 43: this will now happen after reclustering, so the UUID recorded on the event log will be the new cluster if the record has moved

PR #1238, opened 2025-07-22
CPR-779 Add cluster record composition endpoint

H-Iwanejko:


markreesmoj (approved):
Might be an idea to have a separate test (and possibly controller) for the new endpoint but looks good

PR #1246, opened 2025-07-23
FIX: try to make test more reliable - theory is that the order is not…

markreesmoj:
… deterministic

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceIntTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceIntTest.kt
@@ -116,7 +116,7 @@ class ReclusterServiceIntTest : MessagingMultiNodeTestBase() {
 
       stubPersonMatchUpsert()
       stubXPersonMatches(matchId = doesNotMatch.matchId, aboveJoin = listOf(matchesA.matchId, recordA.matchId))
-      stubClusterIsValid(requestBody = """["${recordA.matchId}", "${matchesA.matchId}","${doesNotMatch.matchId}"]""")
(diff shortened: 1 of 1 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceIntTest.kt
```diff
@@ -57,37 +57,7 @@ class ReclusterServiceIntTest : MessagingMultiNodeTestBase() {
     }
 
     @Test
```
markreesmoj, line 59: covered by https://github.com/ministryofjustice/hmpps-person-record/blob/c0a6abc3d6553d868073ba25bf346b5c88794b4f/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceIntTest.kt#L78 as far as I can see

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceIntTest.kt
```diff
@@ -116,7 +86,7 @@ class ReclusterServiceIntTest : MessagingMultiNodeTestBase() {
 
       stubPersonMatchUpsert()
       stubXPersonMatches(matchId = doesNotMatch.matchId, aboveJoin = listOf(matchesA.matchId, recordA.matchId))
-      stubClusterIsValid(requestBody = """["${recordA.matchId}", "${matchesA.matchId}","${doesNotMatch.matchId}"]""")
+      stubClusterIsValid()
```
markreesmoj, line 89: we have coverage of this elsewhere and it seems that the order can be unreliable

PR #1279, opened 2025-08-04
CPR-744 Add nationality tables

H-Iwanejko:


src/main/resources/db/migration/V106__add_nationality_fields.sql
```diff
@@ -0,0 +1,40 @@
+BEGIN;
```
JunaedAhmedMOJ, line 1: using V106 for ethnicity

PR #1288, opened 2025-08-07
CPR-688 Map Nationality Codes

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/types/nationality/CommonPlatformNationalityMappings.kt
```diff
@@ -0,0 +1,194 @@
+package uk.gov.justice.digital.hmpps.personrecord.model.types.nationality
+
+val COMMON_PLATFORM_NATIONALITY_MAPPING: Map<String, NationalityCode> = mapOf(
+  "AF" to NationalityCode.AFGA,
```
samz-justice, line 4: is there a way reduce the repeated NationalityCode?

PR #1305, opened 2025-08-12
CPR-730 ethnicity migration

JunaedAhmedMOJ:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/migrations/MigrateEthnicityCode.kt
```diff
@@ -0,0 +1,87 @@
...
+    log.info(
+      "Finished migrating ethnicity codes, total pages: ${executionResults.totalPages}, " +
+        "total elements: ${executionResults.totalElements}, " +
+        "elapsed time: ${executionResults.elapsedTime}",
+    )
+  }
+
+  private fun PersonEntity.getEthnicityCode(): EthnicityCode? = EthnicityCode.from(this.ethnicity)
+
+  private fun List<EthnicityCodeEntity>.lookupEthnicityCode(ethnicityCode: EthnicityCode?): EthnicityCodeEntity? = ethnicityCode?.let { this.find { it.code == ethnicityCode.name } }
```
markreesmoj, line 54: Maybe it doesn't matter too much, but this would probably be quicker implemented as a map of `name` to `EthnicityCodeEntity`

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/migrations/MigrateEthnicityCode.kt
```diff
@@ -0,0 +1,87 @@
...
+    return OK
+  }
+
+  suspend fun migrateEthnicityCodes() = coroutineScope {
+    log.info("Starting migration of ethnicity codes")
+    val preLoadedEthnicityCodes: List<EthnicityCodeEntity> = ethnicityCodeRepository.findAll()
+    val executionResults = forPage { page ->
+      log.info("Migrating ethnicity codes, page: ${page.pageable.pageNumber + 1}")
+      page.content.map { personEntity ->
+        val ethnicityCode = personEntity.getEthnicityCode()
```
markreesmoj, line 38: I assume that new records and updates to existing records are already storing the new ethnicityCode? 
If so, does this code work with them?

JunaedAhmedMOJ: it will yes but there was an issue with a handful of codes which were mapped as unknown so this wouldnt be so bad to do

markreesmoj: Please can we have a test which proves this behaviour?

PR #1314, opened 2025-08-14
CPR-730 fix unknowns

JunaedAhmedMOJ:


src/main/resources/db/migration/V112__edit_ethnicity_code_descriptions.sql
```diff
@@ -0,0 +1,45 @@
+BEGIN;
+-------------------------------------------------------
+UPDATE ethnicity_codes
+SET description = 'Asian/Asian British : Chinese'
+WHERE id = 4;
```
markreesmoj, line 5: Are these IDs going to be the same across all environments?
Could we use the code instead?

samz-justice (approved):

PR #1328, opened 2025-08-20
CPR-831 Store links

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAPIController.kt
```diff
@@ -37,10 +41,22 @@ class ProbationAPIController(
   fun createProbationRecord(
     @RequestBody probationCase: ProbationCase,
   ) {
-    createUpdateService.processPerson(Person.from(probationCase)) {
-      probationCase.identifiers.crn?.let {
+    when {
+      probationCase.hasMissingCrnOrDefendantId() -> throw MalformedDataException("Missing identifier: CRN / Defendant ID")
+    }
+    probationCase.createPerson()
```
samz-justice, line 47: createPerson will throw if not enough information passed in the body, correct?
so createPerson will raise an error.
therefore, do we really need to throw our own check?
if the 2 piece of information are vital to the call, then perhaps we can inject them as part of the url parameters (then it will be a 404 if neither of them are supplied)

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/IntegrationTestBase.kt
```diff
@@ -598,10 +602,19 @@ class IntegrationTestBase {
       personRepository.findByMatchId(this.matchId)?.overrideMarkers?.filter { it.markerType == EXCLUDE && it.markerValue == personEntity.id },
     ).hasSize(1)
   }
+
   fun PersonEntity.assertPersonDeleted() = awaitAssert { assertThat(personRepository.findByMatchId(this.matchId)).isNull() }
 
   fun PersonKeyEntity.assertPersonKeyDeleted() = awaitAssert { assertThat(personKeyRepository.findByPersonUUID(this.personUUID)).isNull() }
 
+  internal fun String.assertLinksToCrn(crn: String) = awaitAssert {
+    assertThat(courtProbationLinkRepository.findByDefendantId(this)?.crn).isEqualTo(crn)
```
samz-justice, line 611: the call is to findByDefendantId then we compare to crn
shouldn't we either compare to defendantId
or call findByCrn(this) would make it easier to understand?

markreesmoj (approved):
❤️

PR #1329, opened 2025-08-20
CPR-831 Rename Name data class for proper swagger documentation

H-Iwanejko:
* Renamed as swagger evaluates the first `Name` object it finds as the docs so can have incorrect request payload examples with object of the same name


markreesmoj (approved):
👍

markreesmoj (approved):
👍

PR #1333, opened 2025-08-21
CPR-830 Return probation links

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/court/CourtAPIController.kt
```diff
@@ -48,7 +52,13 @@ class CourtAPIController(
...
       personEntity == null -> throw ResourceNotFoundException(defendantID)
-      else -> ResponseEntity.ok(CanonicalRecord.from(personEntity))
+      else -> ResponseEntity.ok(personEntity.buildRecord())
     }
   }
+
+  private fun PersonEntity.buildRecord(): CanonicalRecord {
+    val courtLink = courtProbationLinkRepository.findByDefendantId(this.defendantId!!)
+    val canonicalIdentifiers: CanonicalIdentifiers = CanonicalIdentifiers.from(this, additionalCrn = courtLink?.crn)
+    return CanonicalRecord.from(this, identifiers = canonicalIdentifiers)
```
H-Iwanejko, line 62: I could make the crn list mutable and add it here? if that seems sensible

PR #1351, opened 2025-08-27
CPR-835 Initial new unmerge with override markers

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
```diff
@@ -154,6 +154,11 @@ class PersonEntity(
     )
   }
 
+  fun addOverrideMarker(scope: OverrideScopeEntity) {
+    this.overrideMarker = this.overrideMarker ?: OverrideScopeEntity.newMarker()
```
H-Iwanejko, line 158: Don't override marker as scenario where this record is included i.e. same marker. Should not overwrite it

samz-justice (approved):

PR #1355, opened 2025-08-27
CPR-857 libra title update

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/types/TitleCode.kt
```diff
@@ -23,26 +23,32 @@ enum class TitleCode {
...
+      "SISTER" to SR,
+      "DAME" to DME,
+      "DME" to DME,
+      "DR" to DR,
+      "LDY" to LDY,
+      "LADY" to LDY,
+      "LRD" to LRD,
+      "LORD" to LRD,
+      "SIR" to SIR,
+    )
```
markreesmoj, line 49: to enable test setup with all possible title values

PR #1360, opened 2025-08-28
CPR-836 Migrate excludes to override marker + scopes

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/OverrideMarkerEntity.kt
```diff
@@ -37,4 +37,7 @@ class OverrideMarkerEntity(
 
   @Version
   var version: Int = 0,
+
+  @Column(name = "fk_person_id", insertable = false, updatable = false)
```
markreesmoj, line 41: Is this because another entity manages this relationship?

markreesmoj (approved):
👍

PR #1362, opened 2025-08-29
CPR-857 simplify ethnicity setup

markreesmoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/types/EthnicityCodeTest.kt
```diff
@@ -0,0 +1,186 @@
...
+import uk.gov.justice.digital.hmpps.personrecord.model.types.EthnicityCode.Z1
+import uk.gov.justice.digital.hmpps.personrecord.test.randomCrn
+import uk.gov.justice.digital.hmpps.personrecord.test.randomDate
+import uk.gov.justice.digital.hmpps.personrecord.test.randomDefendantId
+import uk.gov.justice.digital.hmpps.personrecord.test.randomName
+import uk.gov.justice.digital.hmpps.personrecord.test.randomPrisonNumber
+import java.util.stream.Stream
+import uk.gov.justice.digital.hmpps.personrecord.client.model.offender.ProbationCaseName as OffenderName
+
+class EthnicityCodeTest {
```
markreesmoj, line 53: The descriptions are stored in the database and will not change (unless we change them). They are tested by three integration tests on a randomised basis in any case.
This will run much quicker as a unit test which does not initialise the Spring application context and interact with the database

samz-justice (approved):

PR #1366, opened 2025-08-29
CPR-79 Run tests with opts

H-Iwanejko:
* Actually runs tests with java options.
* `Picked up _JAVA_OPTIONS: -Xmx4096m... `

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -49,6 +49,7 @@ jobs:
     uses: ./.github/workflows/gradle_localstack_postgres_verify.yml # WORKFLOW_VERSION
     secrets: inherit
     with:
+      java-options: "${{ inputs.java_options }}"
       postgres-tag: '16.1'
       localstack-tag: '3.0'
       postgres-username: 'root'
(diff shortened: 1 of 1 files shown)


.github/workflows/pipeline.yml
```diff
@@ -48,7 +49,7 @@ jobs:
     uses: ./.github/workflows/gradle_localstack_postgres_verify.yml # WORKFLOW_VERSION
     secrets: inherit
     with:
-      java-options: "${{ inputs.java_options }}"
+      gradle-command: "make test-all"
```
markreesmoj, line 52:
```suggestion
      gradle-command: "export _JAVA_OPTIONS=${JAVA_OPTS} && make test-all"
```
This means we only have to specify the `JAVA_OPTIONS` in the pipeline and will make it easier for people to reuse if we contribute it back. The PR is fine as is, can we do anything to make it more reusable?

H-Iwanejko: Yeah might be worth updating the shared action to do this, gonna update to have a reusable java opts

PR #1369, opened 2025-09-01
CPR-838 Add status reason code to needs attention

H-Iwanejko:


samz-justice (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonKeyEntity.kt
```diff
@@ -51,6 +51,20 @@ class PersonKeyEntity(
 
   fun isNeedsAttention(): Boolean = status == UUIDStatusType.NEEDS_ATTENTION
 
+  fun setAsActive() {
+    this.apply {
+      this.status = UUIDStatusType.ACTIVE
+      this.statusReason = null
+    }
```
markreesmoj, line 58: this could be either

```suggestion
     this.apply {
      status = UUIDStatusType.ACTIVE
      statusReason = null
    }
```

or

```suggestion
      this.status = UUIDStatusType.ACTIVE
      this.statusReason = null
```

I know it is following an existing pattern but either of the above are shorter

PR #1372, opened 2025-09-01
CPR-857 adding addtional nationality to common platform

samz-justice:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/reference/NationalityCodeEntity.kt
```diff
@@ -14,7 +14,7 @@ class NationalityCodeEntity(
   @GeneratedValue(strategy = GenerationType.IDENTITY)
   val id: Long? = null,
 
-  @Column(unique = true, nullable = false)
```
markreesmoj, line 17: https://github.com/ministryofjustice/hmpps-person-record/blob/d270eacb566cd23ff9d1de24be1a3733bc97525a/src/main/resources/db/migration/V106__add_nationality_fields.sql#L20
Covered here

H-Iwanejko (approved):

PR #1378, opened 2025-09-02
CPR-838 map status reason to event log

JunaedAhmedMOJ:


H-Iwanejko (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/eventlog/EventLogServiceIntTest.kt
```diff
@@ -159,4 +162,14 @@ class EventLogServiceIntTest : IntegrationTestBase() {
     assertThat(eventLog.sentenceDates[0]).isEqualTo(LocalDate.of(1980, 1, 1))
     assertThat(eventLog.sentenceDates[1]).isEqualTo(LocalDate.of(1990, 1, 1))
   }
+
+  @Test
```
samz-justice, line 166: it might be a better test to add the assertion to existing code that handle logging when a cluster goes to needs attention, then this test can be deleted 
https://github.com/ministryofjustice/hmpps-person-record/blob/b99796179d175a0c4f2aa02c77092d6aad0162dc/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceIntTest.kt#L1275

JunaedAhmedMOJ: good idea, didnt realise this test was doing event logging asserts

samz-justice (approved):
👍

PR #1379, opened 2025-09-02
CPR-838 fix override deletion error

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
```diff
@@ -70,7 +70,7 @@ class PersonEntity(
   @Column(name = "override_marker")
   var overrideMarker: UUID? = null,
 
-  @ManyToMany(cascade = [ALL], fetch = EAGER)
+  @ManyToMany(fetch = EAGER)
```
H-Iwanejko, line 73: causes cascade to delete scope that is still in use

markreesmoj (approved):
👍

PR #1382, opened 2025-09-02
FIX: remove tests with coverage elsewhere

markreesmoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/jobs/RecordCountReportIntTest.kt
```diff
@@ -14,31 +14,7 @@ class RecordCountReportIntTest : WebTestBase() {
...
-        "NOMIS" to "1",
-        "DELIUS" to "1",
-        "COMMON_PLATFORM" to "1",
-        "LIBRA" to "1",
-      ),
-    )
-  }
-
-  @Test
-  fun `should not include record in telemetry`() {
```
markreesmoj, line 41: this test covers everything covered by the deleted test

PR #1383, opened 2025-09-03
Cpr 838 move tests to e2e

H-Iwanejko:
* More the needs attention tests to e2e

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
@@ -25,21 +25,6 @@ class ReclusterServiceE2ETest : E2ETestBase() {
   inner class ClusterAlreadySetAsNeedsAttention {
 
     @Test
-    fun `should add a created record to a cluster if it is set to need attention`() {
-      val basePersonData = createRandomProbationPersonDetails()
-
-      val personA = createPerson(createProbationPersonFrom(basePersonData))
-      val cluster = createPersonKey(status = NEEDS_ATTENTION, reason = OVERRIDE_CONFLICT)
-        .addPerson(personA)
-
-      val newPersonData = createProbationPersonFrom(basePersonData)
-      probationDomainEventAndResponseSetup(eventType = NEW_OFFENDER_CREATED, ApiResponseSetup.from(newPersonData))
-
-      cluster.assertClusterIsOfSize(2)
-      cluster.assertClusterStatus(NEEDS_ATTENTION, reason = OVERRIDE_CONFLICT)
-    }
-
-    @Test
     fun `should retain needs attention status when a record is updated which continues to match only one record out of 2 in the cluster`() {
       val basePersonData = createRandomProbationPersonDetails()
 


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceIntTest.kt
```diff
@@ -39,45 +39,6 @@ class ReclusterServiceIntTest : MessagingMultiNodeTestBase() {
...
-      stubPersonMatchUpsert()
-      stubClusterIsNotValid()
-      probationDomainEventAndResponseSetup(eventType = OFFENDER_PERSONAL_DETAILS_UPDATED, ApiResponseSetup(crn = recordA.crn))
-
-      cluster.assertClusterIsOfSize(3)
-      cluster.assertClusterStatus(NEEDS_ATTENTION, reason = BROKEN_CLUSTER)
-    }
-
     @Test
     fun `should change from needs attention status to active when the non-matching record is updated to match the other records in the cluster above the join threshold`() {
```
H-Iwanejko, line 43: Leaving the `below fracture` and  `above fracture` threshold tests here as it is hard to reliable reproduce those scores. One to tackle later

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
```diff
@@ -5,18 +5,112 @@ import org.junit.jupiter.api.Test
...
+      val basePersonData = createRandomProbationPersonDetails()
+
+      val personA = createPerson(createProbationPersonFrom(basePersonData))
+      val cluster = createPersonKey(status = NEEDS_ATTENTION, reason = OVERRIDE_CONFLICT)
+        .addPerson(personA)
+
+      val newPersonData = createProbationPersonFrom(basePersonData)
+      probationDomainEventAndResponseSetup(eventType = NEW_OFFENDER_CREATED, ApiResponseSetup.from(newPersonData))
+
+      checkTelemetry(
```
markreesmoj, line 38: do we need this check? it is covered in quite a few other tests

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
```diff
@@ -5,18 +5,112 @@ import org.junit.jupiter.api.Test
...
+        TelemetryEventType.CPR_RECORD_CREATED,
+        mapOf("CRN" to newPersonData.crn, "SOURCE_SYSTEM" to "DELIUS"),
+      )
+
+      cluster.assertClusterIsOfSize(2)
+      cluster.assertClusterStatus(NEEDS_ATTENTION, reason = OVERRIDE_CONFLICT)
+    }
+
+    @Test
+    fun `should retain needs attention status when a record is updated which continues to match another record in the cluster and the cluster remains invalid`() {
```
markreesmoj, line 48: let's be consistent with naming - if we are using `needs attention` let's stick with that instead of invalid
I know these are copy/paste from the original names but we have the chance to improve them now

markreesmoj: in fact we can just drop `and the cluster remains invalid` as `should retain needs attention status` gets the intent across

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
H-Iwanejko: Removed 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
```diff
@@ -5,18 +5,111 @@ import org.junit.jupiter.api.Test
...
+      recordToJoinCluster.personKey?.assertMergedTo(cluster)
+    }
+
+    @Test
+    fun `should not set a cluster to active if it is set to needs attention and an update does change the cluster composition`() {
+      val basePersonData = createRandomProbationPersonDetails()
+
+      val personA = createPerson(createProbationPersonFrom(basePersonData))
+      val cluster = createPersonKey(status = NEEDS_ATTENTION, reason = BROKEN_CLUSTER)
+        .addPerson(personA)
```
markreesmoj, line 93: This might be the case as the code currently stands but it is odd behaviour. 

I think the following is true:
This cluster has 1 record and is NEEDS_ATTENTION. 
Another record joins it
This cluster now consists of 2 records which match and so isClusterValid would return true. Any further update to this cluster which does not reduce the match weight will make it leave NEEDS_ATTENTION 

Arguably this is a bug (although an unlikely scenario). I think there might be the odd cluster with only 1 record in which needs attention. 

I think it would be clearer if the cluster went from 2 records to 3 in this test

H-Iwanejko: Yeah discovered this and fixed it in the new reclustering process:
https://github.com/ministryofjustice/hmpps-person-record/pull/1371/files#diff-8a47e16e6e5773a633ffb5f555a9bd2dd2bb77fac16a10c1cf3f859d26506e00R178
Should self heal a cluster by its self if it is a broken cluster. But this is a valid scenario where a record can join a NEEDS attention cluster

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
```diff
@@ -5,18 +5,111 @@ import org.junit.jupiter.api.Test
...
+
+      val newPersonData = createProbationPersonFrom(basePersonData)
+      probationDomainEventAndResponseSetup(eventType = NEW_OFFENDER_CREATED, ApiResponseSetup.from(newPersonData))
+
+      checkTelemetry(
```
markreesmoj, line 98: :-) this check is not needed

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/match/PersonMatchRecord.kt
```diff
@@ -44,8 +43,7 @@ data class PersonMatchRecord(
       sentenceDates = personEntity.sentenceInfo.mapNotNull { it.sentenceDate }.map { it.toString() }.distinct().sorted(),
       sourceSystemId = personEntity.extractSourceSystemId(),
       overrideMarker = personEntity.overrideMarker?.toString() ?: "",
-      overrideScopes = personEntity.overrideScopes.mapNotNull { overrideScope -> overrideScope.scope.toString() },
-
+      overrideScopes = personEntity.overrideScopes.map { overrideScope -> overrideScope.scope.toString() },
```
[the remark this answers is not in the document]
H-Iwanejko: The `scope` is a non-nullable field so the `NotNull` bit is redundant. linter advisesd

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
markreesmoj: got it, nice one. It's hard to see in this view

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
```diff
@@ -5,18 +5,109 @@ import org.junit.jupiter.api.Test
...
+
+      val personA = createPerson(createProbationPersonFrom(basePersonData))
+      val cluster = createPersonKey(status = NEEDS_ATTENTION, reason = OVERRIDE_CONFLICT)
+        .addPerson(personA)
+
+      val newPersonData = createProbationPersonFrom(basePersonData)
+      probationDomainEventAndResponseSetup(eventType = NEW_OFFENDER_CREATED, ApiResponseSetup.from(newPersonData))
+
+      cluster.assertClusterIsOfSize(2)
+      cluster.assertClusterStatus(NEEDS_ATTENTION, reason = OVERRIDE_CONFLICT)
```
markreesmoj, line 39: this is covered in `should not set a cluster to active if it is set to needs attention and an update does change the cluster composition` so seems unnecessary now

PR #1384, opened 2025-09-03
CPR-838 Add more tests to e2e

H-Iwanejko:
* Move no change + race conditions to e2e

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
@@ -5,9 +5,11 @@ import org.junit.jupiter.api.Test
 import org.springframework.beans.factory.annotation.Autowired
 import uk.gov.justice.digital.hmpps.personrecord.config.E2ETestBase
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.PersonEntity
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
```diff
@@ -237,9 +239,152 @@ class ReclusterServiceE2ETest : E2ETestBase() {
...
+    fun `should not merge an active cluster to a matched cluster marked as merged`() {
+      val basePersonData = createRandomProbationPersonDetails()
+
+      val personA = createPerson(createProbationPersonFrom(basePersonData))
+      val cluster1 = createPersonKey()
+        .addPerson(personA)
+
+      val personB = createPerson(createProbationPersonFrom(basePersonData))
+      val cluster2 = createPersonKey(MERGED)
+        .addPerson(personB)
```
markreesmoj, line 368: this can't happen, can it?
If a cluster is merged, all the person records on it are moved to the cluster it has been merged to?

H-Iwanejko: This is a potential race condition, potential for a record to still be (in the process of) being moved to the merged cluster.

This test is just to make sure we are not 'stealing' records from other clusters that are in the process of merging.

PR #1410, opened 2025-09-18
CPR-839 Create include link to person

H-Iwanejko:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationApiE2ETest.kt
```diff
@@ -229,38 +255,44 @@ class ProbationApiIntTest : WebTestBase() {
...
-      )
+      offender.assertIncluded(defendant)
+    }
+  }
+
+  @Nested
+  inner class ErrorScenarios {
 
+    @Test
+    fun `should return Not Found if defendant does not exist`() {
```
markreesmoj, line 273: OK to test this but it cannot happen realistically as the preceding call to https://hmpps-person-record-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/HMPPS%20Person%20API/getRecord has to succeed and therefore the defendant must exist

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAPIController.kt
```diff
@@ -45,16 +53,24 @@ class ProbationAPIController(
     @PathVariable(name = "defendantId") defendantId: String,
     @RequestBody probationCase: ProbationCase,
   ) {
-    probationCase.createPerson()
+    val defendant: PersonEntity = retrieveDefendant(defendantId)
+    val offender: PersonEntity = probationCase.createPersonWithoutLinkingAndClustering()
```
markreesmoj, line 57: can't this just be allowed to join whatever cluster it wants?
99/100 it will already exist anyway I think

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationApiE2ETest.kt
H-Iwanejko: Yeah though i'd should a test in just in case.

PR #1412, opened 2025-09-18
CPR-839 Slight Refactor person service

H-Iwanejko:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/person/PersonService.kt
```diff
@@ -25,34 +25,20 @@ class PersonService(
   private val publisher: ApplicationEventPublisher,
 ) {
 
-  fun handlePersonCreation(person: Person): PersonEntity {
-    val personEntity = personFactory.create(person)
-    personMatchService.saveToPersonMatch(personEntity)
-    if (person.linkOnCreate) {
-      linkRecordToPersonKey(personEntity)
```
H-Iwanejko, line 32: can refactor the way we trigger a recluster bit here, but i thought it would be too big for this PR

samz-justice (approved):

PR #1421, opened 2025-09-22
Speed up tests

markreesmoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/IntegrationTestBase.kt
```diff
@@ -235,32 +229,26 @@ class IntegrationTestBase {
...
     awaitAssert(function = {
       matchingEvents(eventLogRepository.findAllByEventTypeAndPersonUUIDOrderByEventTimestampDesc(event, cluster) ?: emptyList())
-    }, timeout)
+    })
   }
 
-  private fun awaitAssert(function: () -> Unit, timeout: Long) = await atMost (Duration.ofSeconds(timeout)) untilAsserted function
-
-  internal fun awaitAssert(function: () -> Unit) = awaitAssert(function = function, timeout = 3)
+  internal fun awaitAssert(function: () -> Unit) = await atMost (Duration.ofSeconds(15)) untilAsserted function
```
markreesmoj, line 249: not sure this is a great idea, it might be annoying if a lot of tests fail
There's just one test which takes a while and needs a longer timeout

PR #1444, opened 2025-09-29
CPR-912 do the pagination after we get all the records from the db

samz-justice:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/admin/cluster/ClustersController.kt
```diff
@@ -47,15 +43,29 @@ class ClustersController(
...
+      content = content,
       pagination = Pagination(
-        isLastPage = paginatedClusters.isLast,
-        count = paginatedClusters.numberOfElements,
-        page = paginatedClusters.number + 1,
-        perPage = paginatedClusters.size,
-        totalCount = paginatedClusters.totalElements,
-        totalPages = paginatedClusters.totalPages,
+        isLastPage = isLast,
+        totalPages = totalPages.toInt(),
```
markreesmoj, line 68: The pagination only requires these two properties

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/admin/cluster/ClustersController.kt
```diff
@@ -47,15 +43,35 @@ class ClustersController(
...
+      isLastPage -> (totalClusters % DEFAULT_PAGE_SIZE) + first
+      else -> first + DEFAULT_PAGE_SIZE
+    }
+
+    val content = clusters.subList(first, last)
+
     return PaginatedResponse(
-      content = clusters,
+      content = content,
       pagination = Pagination(
```
markreesmoj, line 72: https://github.com/ministryofjustice/hmpps-person-record-manage-ui/blob/main/server/builders/pagination.builder.ts#L46

markreesmoj:
It was either this or some fiendishly complex SQL. This will put the Delius and Nomis records first in no particular order other than when the cluster was originally created. 
The requirement to order by the date the record entered NEEDS_ATTENTION is not easy to reconcile with this query. That information lives in the event log table and there is no relationship between event log and clusters. It does not seem worth adding one to facilitate this query.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/PersonKeyRepository.kt
```diff
@@ -12,5 +10,5 @@ import java.util.*
 interface PersonKeyRepository : JpaRepository<PersonKeyEntity, Long> {
   fun findByPersonUUID(personUUID: UUID?): PersonKeyEntity?
 
-  fun findAllByStatusOrderById(uuidStatus: UUIDStatusType, pageable: Pageable): Page<PersonKeyEntity>
+  fun findAllByStatus(uuidStatus: UUIDStatusType): List<PersonKeyEntity>
```
markreesmoj, line 13: please add Id back in

H-Iwanejko (approved):

PR #1458, opened 2025-10-01
CPR-914 retries

markreesmoj:
https://github.com/spring-projects/spring-retry/issues/180

The retryable annotation only works if you call it from another class. So we are only retrying court messages using retryable at the moment. This will make it work for everything.

Other messages were being retried using the redrive policy on the queue, presumably.


H-Iwanejko (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationEventListenerIntTest.kt
```diff
@@ -270,7 +270,7 @@ class ProbationEventListenerIntTest : MessagingMultiNodeTestBase() {
     fun `test multiple requests to probation single record process successfully`() {
       val pnc = randomPnc()
       val crn = randomCrn()
-      blitz(30, 2) {
+      blitz(30, 15) {
```
markreesmoj, line 273: 🔨

H-Iwanejko (approved):

PR #1460, opened 2025-10-02
CPR-877 retry is in but test failing for some reason

markreesmoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/admin/AdminReclusterService.kt
```diff
@@ -0,0 +1,85 @@
...
+  private val publisher: ApplicationEventPublisher,
+) {
+
+  @Retryable(
+    maxAttempts = 2,
+    backoff = Backoff(delay = 200, random = true, multiplier = 3.0),
+    retryFor = [
+      WebClientException::class,
+    ],
+  )
```
markreesmoj, line 32: not sure this is the right place to retry - it will retry everything from the start I think

PR #1476, opened 2025-10-08
CPR-736 Rename country to code

H-Iwanejko:


src/main/resources/db/migration/V125__rename_country_fields.sql
```diff
@@ -0,0 +1,11 @@
+BEGIN;
```
JunaedAhmedMOJ, line 1: can we do this in a seperate PR to avoid logs getting errors building up, since the updated migration won't persist on existing pods with old code.

PR #1899, opened 2026-03-20
FIX: merge records in tests using merge events instead of mimicking it

markreesmoj:
The original code attempted to mimic what happens when records are merged. However it only did part of the job, specifically it did not:
- remove the personKey of the merged from record
- merge the UUIDs if the merged from record was the only record on the cluster

Using the mergeService to marge the records for test setup gives more realistic setup.
It appears that it has also uncovered a bug.

Given:
KeyA has one record, recordA
Keyb has one record, recordB

When:
RecordA is merged to recordB 

Then: 
RecordA joins keyB and keyA is merged to keyB

When: 
recordB is deleted

Then: 
RecordA is deleted
keyB is deleted
**keyA should be deleted but it is not**


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationDeleteListenerIntTest.kt
```diff
@@ -223,23 +221,18 @@ class ProbationDeleteListenerIntTest : MessagingMultiNodeTestBase() {
...
     checkEventLogExist(recordACrn, CPRLogEvents.CPR_UUID_DELETED)
     checkEventLogExist(recordBCrn, CPRLogEvents.CPR_RECORD_DELETED)
-    checkEventLogExist(recordBCrn, CPRLogEvents.CPR_UUID_DELETED)
     checkEventLogExist(recordCCrn, CPRLogEvents.CPR_RECORD_DELETED)
 
     recordA.assertPersonDeleted()
     recordB.assertPersonDeleted()
     recordC.assertPersonDeleted()
     recordA.personKey?.assertPersonKeyDeleted()
-    recordB.personKey?.assertPersonKeyDeleted()
```
markreesmoj: recordB has no personKey because it has been merged into recordA, and merged records have their personKey link removed: https://github.com/ministryofjustice/hmpps-person-record/blob/965fe9a700069f5a67cd7fd9c27cb6437467ba61/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/MergeService.kt#L43

jldparker527 (approved):

PR #1940, opened 2026-04-09
Made MergeService a little more understandable

jldparker527:
Added javadoc to mergeService

Renamed MergeService to PersonMergeService

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonMergeService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonMergeService.kt
@@ -12,8 +12,20 @@ import uk.gov.justice.digital.hmpps.personrecord.service.cprdomainevents.events.
 import uk.gov.justice.digital.hmpps.personrecord.service.cprdomainevents.events.merge.PersonMerged
 import uk.gov.justice.digital.hmpps.personrecord.service.search.PersonMatchService
 
+/**
+ * Merges one PersonEntity into another PersonEntity.
+ *
+ * This is accomplished by updating the 'mergedTo' column on the FROM PersonEntity
+ * to the primary key of the TO PersonEntity. Lastly the FROM PersonEntity is moved from its original cluster to its new cluster (the cluster belonging to the TO PersonEntity).
+ *
+ * Note: Alongside the primary behavior above, this class also handles some additional work (◞‸◟,).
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonMergeService.kt
```diff
@@ -12,22 +12,35 @@ import uk.gov.justice.digital.hmpps.personrecord.service.cprdomainevents.events.
 import uk.gov.justice.digital.hmpps.personrecord.service.cprdomainevents.events.merge.PersonMerged
 import uk.gov.justice.digital.hmpps.personrecord.service.search.PersonMatchService
 
+/**
+ * Merges one PersonEntity into another PersonEntity.
+ *
+ * This is accomplished by updating the 'mergedTo' column on the FROM PersonEntity
+ * to the primary key of the TO PersonEntity. Lastly the FROM PersonEntity is moved from its original cluster to its new cluster (the cluster belonging to the TO PersonEntity).
```
markreesmoj, line 19: this is not true - the from personEntity has no cluster once finished

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonMergeService.kt
```diff
@@ -12,22 +12,35 @@ import uk.gov.justice.digital.hmpps.personrecord.service.cprdomainevents.events.
...
+ *
+ * Note: Alongside the primary behavior above, this class also handles some additional work (◞‸◟,).
+ * 1. In the event of a from record being merged that belonged to a single node cluster, it will mark the original cluster it belonged to as merged
+ * and point to the cluster the TO PersonEntity belongs to.
```
markreesmoj, line 23: it does but we think this is unnecessary (as per the hack day on Tuesday) and is likely to be removed imminently so not sure it is wise to document here

markreesmoj:
I feel it's better wherever possible to rename and refactor in order to bring clarity to code rather than write comments. We don't have any JavaDoc comments elsewhere and I'm not sure we should start here

markreesmoj:
I like the renaming and refactoring but not sure how well it will play with the in progress work from Tuesday? If no conflicts then happy to go with it

markreesmoj:
https://dsdmoj.atlassian.net/wiki/spaces/PRD/pages/5681086549/Merge+Process+Flow gives a lot of detail on how merges work

PR #1944, opened 2026-04-10
CPR-1140

markreesmoj:
This is just a spike to show how much work is required if we decide to save the address on its own instead of using personService


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/AdditionalInformation.kt
```diff
@@ -20,4 +20,9 @@ data class AdditionalInformation(
 
   @JsonProperty("removedNomsNumber")
   val sourcePrisonNumber: String? = null,
+
+  val addressId: String? = null,
+
+  val cprAddressId: String? = null,
+  val deliusAddressId: String? = null,
```
markreesmoj, line 27: This is confusing - we should consider using a separate AdditionalInformation class for each different domain event

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/CorePersonRecordAndDeliusClient.kt
```diff
@@ -28,4 +29,14 @@ class CorePersonRecordAndDeliusClient(private val corePersonRecordAndDeliusWebCl
...
     .bodyToMono(ProbationCase::class.java)
+
+  fun getAddress(addressId: String?): ProbationAddress? {
+    val address = corePersonRecordAndDeliusWebClient
+      .get()
+      .uri("/person/address/{id}", addressId)
+      .retrieve()
+      .bodyToMono(ProbationAddress::class.java)
+      .block()
+    return address
```
markreesmoj, line 40: this endpoint does not yet exist

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationEventListener.kt
```diff
@@ -1,25 +1,78 @@
...
+        val addressEntity = probationAddress?.let {
+          val coreAddress = Address.from(it)
+          val addressEntity = coreAddress?.let { AddressEntity.from(coreAddress) }
+          addressEntity!!.person = person
+          addressEntity.usages.forEach { usage -> usage.address = addressEntity }
+          addressEntity.contacts.forEach { contactEntity -> contactEntity.address = addressEntity }
+          addressEntity
+        }!!
+
+        addressRepository.save(addressEntity)
```
markreesmoj, line 58: Additionally we need to save the person to hmpps-person-match and recluster

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/type/DomainEventType.kt
```diff
@@ -11,3 +11,5 @@ const val OFFENDER_MERGED = "probation-case.merge.completed"
 const val OFFENDER_UNMERGED = "probation-case.unmerge.completed"
 const val OFFENDER_GDPR_DELETION = "probation-case.deleted.gdpr"
 const val OFFENDER_DELETION = "probation-case.engagement.deleted"
+
+const val OFFENDER_ADDRESS_CREATED = "probation-case.address.created"
```
markreesmoj, line 15: consider renaming these constants to match the event names

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationEventListener.kt
markreesmoj, line 58: We need to save the delius addressId somewhere too so we can handle updates from Delius correctly

lbevan, line 58: Could we have some sort of `AddressService` that serves a similar role as `PersonService`? i.e. deals with creating or updating an address and raising events to subsequently push domain events.

PR #1947, opened 2026-04-13
CPR-1135 Publish CPR person created domain events

lbevan:


lbevan:
Classes such as `PersonDomainEvent` and `PersonDomainEventPublisher` are built to be reused for publishing CPR domain events (we can rename if we think its confusing).
`PersonDomainEventListener`  currently has an event listener for `PersonCreated` with the expectation to add the updated and delete events. We could also consider separate files for these if we want.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/PersonDomainEvent.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty
+
+data class PersonDomainEvent(
+  @JsonProperty("eventType") val eventType: String,
+  @JsonProperty("version") val version: Int = 1,
+  @JsonProperty("description") val description: String,
+  @JsonProperty("detailUrl") val detailUrl: String? = null,
+  @JsonProperty("occurredAt") val occurredAt: String,
+  @JsonProperty("additionalInformation") val additionalInformation: Map<String, String>? = null,
```
jldparker527, line 11: Could we make use of the existing `AdditionalInformation` class here?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/listeners/PersonDomainEventListener.kt
```diff
@@ -0,0 +1,67 @@
...
+import java.time.Instant
+
+@Component
+class PersonDomainEventListener(
+  private val personDomainEventPublisher: PersonDomainEventPublisher,
+  @Value("\${core-person-record.base-url}") private val baseUrl: String,
+) {
+
+  @EventListener
+  fun onPersonCreated(personCreated: PersonCreated) {
```
jldparker527, line 32: Do we want multiple personCreated events captured in multiple places. Would it be better to update the existing `PersonEventListener.onPersonCreated` function to instead?

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/WebTestBase.kt
```diff
@@ -27,6 +27,13 @@ abstract class WebTestBase : WebTestSetup() {
           mockClient,
         ),
       )
+      whenever(hmppsQueueService.findByTopicId("domainevents")).thenReturn(
```
jldparker527, line 41: Do we need to mock this?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/PersonDomainEvent.kt
lbevan: Not sure we need to in this specific case, given there isn't any additional info to send; using the existing would be a little confusing. Mark references something similar here https://github.com/ministryofjustice/hmpps-person-record/pull/1944#discussion_r3065134334 about creating separate info objects for each type. Not sure that is needed either if we can just use a map?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/listeners/PersonDomainEventListener.kt
lbevan: In my opinion the existing pattern is not particularly great. Publishing new internal events that just get consumed downstream creates more events for the app to handle without any additional benefit. 
If we avoid creating events and just do the work in one place then I worry it becomes messy. A few articles suggest separating concerns into separate listeners. The only real need to consider otherwise is if there is a dependency on other work, but so far we don't have any.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/WebTestBase.kt
lbevan: As far as I am aware yes, but happy to change if its not. In fact this is not working currently as one of the web tests still fails to publish to the topic despite it being a mock. Need to rethink.

lbevan: This is specific to web tests where sqs config is turned off.
The issue with the failing test should now be fixed.

src/main/resources/application.yml
```diff
@@ -99,6 +99,9 @@ management:
 retry:
   timeout: 10000
 
+core-person-record:
+  base-url: ${CORE_PERSON_RECORD_BASE_URL:http://localhost:8080}
+
```
markreesmoj, line 104: we haven't done this for the other URLs. It might be a good pattern, if so let's do it for all of them, might save some space in `application-test.xml` and `application-local.xml`

lbevan:
Consider dev only feature flag for now?

jldparker527 (approved):

PR #1948, opened 2026-04-14
CPR-1125 record from and to UUIDs on merge telemetry

markreesmoj:


lbevan (dismissed):
👍

lbevan:
Tests are failing because we expect the event log to show a null from UUID, but moving the generation of the cluster detail means the from UUID is recorded.

lbevan:
> Tests are failing because we expect the event log to show a null from UUID, but moving the generation of the cluster detail means the from UUID is recorded.

Changed the asserts to now check for the UUID even if its been deleted, and added additional checks on those tests where appropriate.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationMergeEventListenerE2ETest.kt
```diff
@@ -103,7 +103,7 @@ class ProbationMergeEventListenerE2ETest : E2ETestBase() {
     checkEventLog(sourceCrn, CPRLogEvents.CPR_RECORD_MERGED) { eventLogs ->
       assertThat(eventLogs).hasSize(1)
       assertThat(eventLogs.first().recordMergedTo).isEqualTo(targetPerson.id)
-      assertThat(eventLogs.first().personUUID).isNull()
```
markreesmoj, line 106: 👍

jldparker527 (approved):

PR #1954, opened 2026-04-16
FIX: Amend wobbly test with better assertion for NI numbers

lbevan:


lbevan:
Order shouldn't matter here, they just get saved as references regardless if they are additional.

jldparker527 (approved):

PR #1957, opened 2026-04-16
CPR-1154 Remove redirect from CanonicalAPIController

lbevan:
We no longer merge clusters so this cannot happen. We will instead return a 404 when the UUID is not found.


lbevan:
Waiting on database prune.

markreesmoj (approved):
👍 Nice tidy up

PR #1961, opened 2026-04-20
CPR-1099 Add load test pipeline

kaustav-surai:
- Adding load test of core-person-record endpoints in pipeline. 
- Tools used - Gatling/Kotlin
- It supports execution in local and also from pipeline


.github/workflows/gatling_run.yml
```diff
@@ -0,0 +1,87 @@
+name: Gatling performance test run
+
+on:
+  push:
+    branches: [ "cpr-load-test" ]
```
lbevan: Can remove this now that it works. I assume we'll just be running this pipeline manually going forward.

.github/workflows/gatling_run.yml
```diff
@@ -0,0 +1,87 @@
...
+      contents: read
+      actions: read
+    timeout-minutes: 60
+    env:
+      CLIENT_ID: ${{ secrets.GATLING_CLIENT_ID }}
+      CLIENT_SECRET: ${{ secrets.GATLING_CLIENT_SECRET }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```
lbevan: Specify the exact version here e.g. `# v6.0.2` so we know current vs new versions on update. Same applies to other actions in this file.

.github/workflows/gatling_run.yml
kaustav-surai: Removed

.github/workflows/gatling_run.yml
kaustav-surai: Updated with the exact version

lbevan:
@kaustav-surai We've had a chat and we think separating the gatling code out to a separate repository is best.

PR #1965, opened 2026-04-21
CPR-1156 Add status_code to address table

lbevan:


lbevan:
Trying to be consistent with enum naming using `status_code`.

jldparker527 (approved):

PR #1968, opened 2026-04-22
CPR-1138 remove check for merged records

markreesmoj:
which will not be in hmpps-person-match anyway and will not have a UUID either
 
 https://github.com/ministryofjustice/hmpps-person-record/blob/aacc6c6cf60e753527647ccecb5b50d2f529e58d/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/search/PersonMatchService.kt#L42 will already have removed any merged records which somehow remained in `hmpps-person-match` (which should not be possible anyway)


lbevan (approved):
👍

PR #1969, opened 2026-04-22
CPR-1156 Map status_code address column

lbevan:


lbevan:
Includes mapping from syscon address. Probation mapping is under a different ticket.

jldparker527 (approved):

PR #1971, opened 2026-04-22
CPR-1156 Remove mapping for primary and mail address columns

lbevan:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/AddressEntity.kt
```diff
@@ -105,12 +105,6 @@ class AddressEntity(
   @Column(name = "record_type")
   var recordType: AddressRecordType? = null,
 
-  @Column(name = "\"primary\"")
-  val primary: Boolean? = null,
```
markreesmoj, line 109: love to see this gone, terrible name for a column. Looks like a reserved word in SQL

markreesmoj (approved):
👍

PR #1972, opened 2026-04-22
FIX: apply owasp update which has been out for 10 days

markreesmoj:
Splitting this out from the more complicated springdoc update

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,13 +1,5 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": ["github>ministryofjustice/hmpps-renovate-config:jvm"],
-  "prBodyTemplate": "{{{table}}}{{{notes}}}{{{warnings}}}{{{controls}}}",
-  "packageRules": [
-    {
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -1,13 +1,5 @@
...
   "extends": ["github>ministryofjustice/hmpps-renovate-config:jvm"],
-  "prBodyTemplate": "{{{table}}}{{{notes}}}{{{warnings}}}{{{controls}}}",
-  "packageRules": [
-    {
-      "matchManagers": ["gradle"],
-      "matchUpdateTypes": ["minor", "patch"],
-      "groupName": "all non major Gradle dependencies",
-      "groupSlug": "all-gradle-minor-patch"
-    }
-  ]
```
markreesmoj, line 12: given the security climate I think we are better off with individual PRs for each dependency update for now. The only exception is when a group of Kotlin dependencies are updated to the latest Kotlin version, and even then we are probably better waiting until the hmpps-gradle-spring-boot plugin has been updated first

lbevan (approved):
👍

PR #1976, opened 2026-04-22
CPR-1156 Drop primary and mail columns from address table

lbevan:


lbevan:
There is no data for these columns so no migration is needed.

markreesmoj (approved):
Provided that the PR which removes the references to these fields from the entities goes to production before this is merged to avoid any errors during deployment then 👍

PR #1977, opened 2026-04-22
FIX: fixing warnings which appear on each build

markreesmoj:
w: file:///home/runner/work/hmpps-person-record/hmpps-person-record/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/admin/ReclusterController.kt:60:5 'when' is exhaustive so 'else' is redundant here.

w: file:///home/runner/work/hmpps-person-record/hmpps-person-record/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/person/Person.kt:229:34 Unnecessary safe call on a non-null receiver of type 'List<Identifier>'.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/person/Person.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/person/Person.kt
@@ -67,8 +67,6 @@ data class Person(
 
   companion object {
 
-    fun List<Reference>.toString(): String = this.joinToString { it.identifierValue.toString() }
-
     fun from(probationCase: ProbationCase): Person {
       val contacts: List<Contact> = listOfNotNull(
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/person/Person.kt
```diff
@@ -67,8 +67,6 @@ data class Person(
 
   companion object {
 
-    fun List<Reference>.toString(): String = this.joinToString { it.identifierValue.toString() }
-
```
markreesmoj, line 71: Appears to have been added in this monster PR for testing purposes: https://github.com/ministryofjustice/hmpps-person-record/commit/72b37a5d58563300cd6dae85c2c9478dcc0b2cbd

lbevan (approved):
👍

PR #1979, opened 2026-04-23
CPR 951 Updated PersonDeletionService to trigger re-clusters on remaining persons in cluster

jldparker527:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/events/person/PersonDeleted.kt
```diff
@@ -2,5 +2,10 @@ package uk.gov.justice.digital.hmpps.personrecord.service.cprdomainevents.events
 
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.PersonEntity
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.PersonKeyEntity
+import java.util.UUID
 
-data class PersonDeleted(val personEntity: PersonEntity, val cluster: PersonKeyEntity?)
+data class PersonDeleted(
+  val personEntity: PersonEntity,
+  val cluster: PersonKeyEntity?,
+  val uuidOfOverrideCluster: UUID? = null,
```
markreesmoj: this is not the UUID of the override exclusion cluster. AFAIK you cannot override exclude a cluster

it is the override marker for the record which has been deleted

Maybe just call it overrideMarker?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonDeletionService.kt
```diff
@@ -48,6 +56,11 @@ class PersonDeletionService(
   }
 
   private fun PersonEntity.deletePersonEntityThatWasMergedIntoThisOneRecursively() {
-    personRepository.findByMergedTo(this.id!!).forEach { personEntity -> processDelete { personEntity } }
+    personRepository.findByMergedTo(this.id!!).filterNotNull().forEach { personEntity -> deletePerson(personEntity) }
```
markreesmoj: Could we make the list nullable instead here?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonDeletionService.kt
```diff
@@ -48,6 +56,11 @@ class PersonDeletionService(
...
+  }
+
+  private fun triggerReclusterOfRemainingNonMergedPersonsInCluster(deletedPersonEntity: PersonEntity, cluster: PersonKeyEntity?) {
```
markreesmoj: name is a but much
`reclusterRemaingPeopleInCluster`?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonDeletionService.kt
```diff
@@ -48,6 +56,11 @@ class PersonDeletionService(
...
+    val remainingNonMergedPersonsInCluster = cluster?.personEntities?.filter { it.mergedTo == null && deletedPersonEntity.id != it.id } ?: emptyList()
```
markreesmoj: there cannot be any merged people in the cluster can there?

markreesmoj: if we have deleted the person by this point we do not need to check that either

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/events/person/PersonDeleted.kt
lbevan: I dont think a new var here is needed? We should be able to get the overrideMarker from the personEntity. See next comment for usage in telementry publish.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/listeners/PersonEventListener.kt
```diff
@@ -45,6 +46,7 @@ class PersonEventListener(
         personDeleted.personEntity,
         mapOf(
           UUID to personDeleted.cluster?.personUUID?.toString(),
+          UUID_OF_OVERRIDE_CLUSTER to personDeleted.uuidOfOverrideCluster.toString(),
```
lbevan: As per my previous comment, can we do `personDeleted.personEntity.overrideMarker.toString()`  instead?

jldparker527: Yes that would be much cleaner

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonDeletionService.kt
jldparker527: We could but we would have to wrap it with a let

`personRepository.findByMergedTo(this.id!!).forEach { personEntity -> personEntity?.let { deletePerson(it) } }`

I think using filterNotNull is more declarative

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonDeletionService.kt
jldparker527: agree

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonDeletionService.kt
jldparker527: No, from persons would not be in the cluster at this point. I chose the name to be extra declarative however, I can see how mentioned merged people in the name could be counter intuitive.

As for the `it.mergedTo == null` check, it serves as a guard rail.. We should decide if we should keep guard rails in place as a general rule moving forward

jldparker527: If we choose to scrap guard rails, we could also remove the `deleteFromPersonMatch` call when merged records are being passed in as they should have been delete from personMatch already

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/events/person/PersonDeleted.kt
jldparker527: 👍

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
```diff
@@ -163,11 +163,6 @@ class ReclusterServiceE2ETest : E2ETestBase() {
         .addPerson(recordToDelete)
       publishProbationDomainEvent(OFFENDER_DELETION, recordToDelete.crn!!)
 
-      cluster.assertClusterIsOfSize(1)
-      cluster.assertClusterStatus(NEEDS_ATTENTION, reason = BROKEN_CLUSTER)
-
```
jldparker527, line 168: Originally re-cluster behaviour was not triggered via delete so we can remove the preceding event to process person

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonDeletionService.kt
markreesmoj: I know we have been here before and this has not changed in this PR but all this recursion needs to do is 

```
personEntity.delete()
personEntity.deletePersonEntityThatWasMergedIntoThisOneRecursively()
```

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/ReclusterServiceE2ETest.kt
markreesmoj: excellent, nice to get rid of this setup

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/TelemetryService.kt
```diff
@@ -42,6 +42,7 @@ enum class EventKeys {
   UUID_COUNT,
   UUID,
   CLUSTER_SIZE,
+  UUID_OF_OVERRIDE_CLUSTER,
```
markreesmoj: this name does not describe what it contains very well. it is the override marker on the person entity being deleted. It is not related to a cluster 
Do we need anything more than a flag stating that a person with an override marker has been deleted? 
What is this actually for? I cannot see what use this UUID will be. The CRN/Prison Number in this event will identify the Person. This is just another piece of data from it

Would be good to understand how this will be used

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonDeletionService.kt
jldparker527: Are we saying then we are happy to remove the guard rail for the call to DELETE personMatch for from records then

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonDeletionService.kt
jldparker527: removed

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/message/PersonDeletionService.kt
```diff
@@ -1,14 +1,15 @@
 package uk.gov.justice.digital.hmpps.personrecord.service.message
 
-import jakarta.transaction.Transactional
 import org.springframework.context.ApplicationEventPublisher
 import org.springframework.stereotype.Component
+import org.springframework.transaction.annotation.Transactional
```
markreesmoj, line 5: ❤️

markreesmoj (approved):
👍

PR #1981, opened 2026-04-23
CPR-1143 Cleanup and organise swagger docs

lbevan:


lbevan:
Proposing we group by source system for now. Helps given the additional granular endpoints we are adding for each source system.

markreesmoj (approved):
❤️

PR #1984, opened 2026-04-23
CPR-1143 Add latest address fields to CanonicalAddress

lbevan:


lbevan:
These fields are needed for CPR-1143 but I think it makes sense to standardise these types of endpoints to use Canonical representations. There may be a few exceptions where we add some additional fields, **but should these new ones be part of the canonical record? I think most (if not all) are used by the source systems. Input appreciated.**

Lots of test changes to add these fields but they are largely the same across the GET endpoints. Seems like a good area to refactor, just difficult to see as part of this.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
```diff
@@ -26,10 +26,16 @@ data class CanonicalAddress(
   val postTown: String? = null,
   @Schema(description = "Person address county", example = "Greater London")
   val county: String? = null,
-  @Schema(description = "Person address country", example = "United Kingdom")
-  val country: String? = null,
+  @Schema(description = "Person address country")
+  val country: CanonicalCountry,
   @Schema(description = "Person address uprn", example = "100120991537")
   val uprn: String? = null,
+  @Schema(description = "Person address status", example = "United Kingdom")
```
markreesmoj: Looks like these 2 examples need updating

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
```diff
@@ -26,10 +26,16 @@ data class CanonicalAddress(
   val postTown: String? = null,
   @Schema(description = "Person address county", example = "Greater London")
   val county: String? = null,
-  @Schema(description = "Person address country", example = "United Kingdom")
-  val country: String? = null,
+  @Schema(description = "Person address country")
```
markreesmoj, line 29: What does the CanonicalCountry example look like?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
```diff
@@ -43,7 +49,16 @@ data class CanonicalAddress(
...
+      country = CanonicalCountry.from(addressEntity.countryCode),
+      uprn = addressEntity.uprn,
+      status = CanonicalAddressStatus.from(addressEntity.statusCode),
+      comment = addressEntity.comment,
+      usages = getUsages(addressEntity),
     )
+
     fun fromAddressEntityList(addressEntity: List<AddressEntity>): List<CanonicalAddress> = addressEntity.map { from(it) }
+
+    private fun getUsages(address: AddressEntity): List<CanonicalAddressUsage> = address.usages.let { CanonicalAddressUsage.fromAddressUsageEntityList(it) }
```
markreesmoj: this function isn't doing a lot - could simply be inlined as `CanonicalAddressUsage.fromAddressUsageEntityList(address.usages)`

markreesmoj:
I think this is OK, worth talking to Marius/Tony to see what they think about it

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
lbevan: Removed the example.
On other schemas we don't provide examples for composite objects. We _could_ provide a json string representation, but I think that would be a lot of overhead as we would need to keep an eye on the internal structure changes and manually change.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
lbevan: On other schemas we don't provide examples for composite objects. We could provide a json string representation, but I think that would be a lot of overhead as we would need to keep an eye on the internal structure changes and manually change.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
lbevan: Inlined 👍

lbevan:
Team has [confirmed](https://mojdt.slack.com/archives/C054W4V32R2/p1777023928536309) we are OK to add these fields to the `CanonicalAddress`.

markreesmoj (approved):
👍

PR #1989, opened 2026-04-27
CPR 1140 saves address events when consuming from probation

jldparker527:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/listeners/AddressCreatedListener.kt
```diff
@@ -0,0 +1,37 @@
...
+import uk.gov.justice.digital.hmpps.personrecord.client.model.sqs.messages.domainevent.DomainEvent
+import uk.gov.justice.digital.hmpps.personrecord.client.model.sqs.messages.domainevent.PersonIdentifier
+import uk.gov.justice.digital.hmpps.personrecord.client.model.sqs.messages.domainevent.PersonReference
+import uk.gov.justice.digital.hmpps.personrecord.extensions.nowUtcFormattedUk
+import uk.gov.justice.digital.hmpps.personrecord.service.cprdomainevents.events.AddressCreated
+import uk.gov.justice.digital.hmpps.personrecord.service.queue.DomainEventPublisher
+
+@Component
+@Profile("!preprod & !prod")
+class AddressCreatedListener(
```
jldparker527, line 19: Not sure if we need another layer of abstraction here. Could simply just call domainEventPublisher.publish from `ProbationEventListener`

jldparker527:
Got some stashed additions that splits out the logic in the ProbationEventListener class but PR is already getting big. Will put in another PR after this is merged in

jldparker527:
awaiting exact callback url from probation. Do not close

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/listeners/AddressCreatedListener.kt
```diff
@@ -0,0 +1,37 @@
...
+  private val domainEventPublisher: DomainEventPublisher,
+  @Value("\${core-person-record.base-url}") private val baseUrl: String,
+) {
+
+  @EventListener
+  @TransactionalEventListener
+  fun onAddressCreated(addressCreated: AddressCreated) {
+    val addressEntity = addressCreated.addressEntity
+    val domainEvent = DomainEvent(
+      "core-person-record.probation.address.created",
```
jldparker527: This may need to be configurable per source system

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/listeners/AddressCreatedListener.kt
```diff
@@ -0,0 +1,37 @@
...
+
+@Component
+@Profile("!preprod & !prod")
+class AddressCreatedListener(
+  private val domainEventPublisher: DomainEventPublisher,
+  @Value("\${core-person-record.base-url}") private val baseUrl: String,
+) {
+
+  @EventListener
+  @TransactionalEventListener
```
lbevan, line 25: Only need `@TransactionalEventListener` here

src/test/resources/application-test.yml
```diff
@@ -40,6 +44,11 @@ hmpps.sqs:
       subscribeTopicId: cprcourtcasestopic
       dlqName: test_only_cpr_court_cases_queue_dlq
       dlqMaxReceiveCount: 2
+    testcpreventsqueue:
+      queueName: test_only_cpr_events_queue
+      subscribeTopicId: domainevents
+      dlqName: test_only_cpr_events_queue_dlq
+      subscribeFilter: '{"eventType":[ "core-person-record.probation.address.created"] }'
```
lbevan: I think you can just use the `testcprdomaineventsqueue` that has been setup previously, adding the new event type to the `subscribeFilter`.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/listeners/AddressCreatedListener.kt
lbevan: It will be eventually. May need something similar to what `PersonDomainEventListener` does.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/MessagingTestBase.kt
```diff
@@ -46,6 +46,10 @@ abstract class MessagingTestBase : IntegrationTestBase() {
     hmppsQueueService.findByQueueId("testcourtcasesqueue")
   }
 
+  val testOnlyCPREventsQueue by lazy {
+    hmppsQueueService.findByQueueId("testcpreventsqueue")
+  }
+
```
lbevan: Should be able to reuse the `testOnlyCPRDomainEventsQueue` below.

src/test/resources/application-test.yml
```diff
@@ -23,6 +23,10 @@ service-now:
   requestor: abcd
   requested-for: efgh
 
+core-person-record:
+  base-url: http://localhost:8080
+
+
```
lbevan: Shouldn't need this. Its already set up in `application.yml`

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/listeners/AddressCreatedListener.kt
lbevan: I think its good if we have the service area publish the AddressCreated/Updated/Deleted event, that then gives us flexibility to add functionality on top. And then these listeners can listen to that.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/events/AddressCreated.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.personrecord.service.cprdomainevents.events
+
+import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.AddressEntity
+
+data class AddressCreated(
+  val crn: String,
+  val addressId: String?,
```
lbevan: Would suggest renaming as I first thought it was our address Id rather than delius.
This is only for short term while we hold this Id, eventually PI will hold it.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/listeners/AddressCreatedListener.kt
jldparker527: When I remove EventListener, the `publisher.publishEvent(AddressCreated(crn, deliusAddressId, addressEntity))` called does not get picked up

src/test/resources/application-test.yml
jldparker527: Good shout 👍

src/test/resources/application-test.yml
jldparker527: removed

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/events/AddressCreated.kt
jldparker527: yes, renamed to `externalAddressId`

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/offender/ProbationAddress.kt
```diff
@@ -1,7 +1,9 @@
 package uk.gov.justice.digital.hmpps.personrecord.client.model.offender
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import java.time.LocalDate
 
+@JsonIgnoreProperties(ignoreUnknown = true)
 data class ProbationAddress(
```
jldparker527, line 7: Little apprehensive about adding these fields to this class as it's used in the probationApi controller and when consuming person events from probation. However, have added `JsonIgnoreProperties` which shouldn't break existing flows

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/AdditionalInformation.kt
```diff
@@ -20,4 +20,8 @@ data class AdditionalInformation(
 
   @JsonProperty("removedNomsNumber")
   val sourcePrisonNumber: String? = null,
+
+  val cprAddressId: String? = null,
+  val addressId: String? = null,
```
jldparker527, line 25: This is the source systems specific addressId! not sure what the actual name of the key will be from the event

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/person/Address.kt
```diff
@@ -66,6 +66,8 @@ data class Address(
       postTown = address.townCity.nullIfBlank(),
       county = address.county.nullIfBlank(),
       uprn = address.uprn.nullIfBlank(),
+      statusCode = address.status?.code?.let { AddressStatusCode.valueOf(it) },
+      recordType = address.type?.code?.let { AddressRecordType.valueOf(it) },
       comment = address.notes.nullIfBlank(),
```
jldparker527, line 71: Will they be sending the same codes here?

jldparker527: Swagger does doesn't show us the enum range

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/offender/ProbationAddress.kt
```diff
@@ -17,4 +19,18 @@ data class ProbationAddress(
   val uprn: String? = null,
   val notes: String? = null,
   val telephoneNumber: String? = null,
+  val status: ProbationAddressStatus? = null,
+  val type: ProbationAddressType? = null,
```
jldparker527, line 23: Not sure if we even need to set these. All DELIUS address have null status and record types in prod

jldparker527: even though the `/probation-cases/{id}` for probationCases contains the fields in their response body

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/person/Address.kt
jldparker527: There not sending them at all even though it's in their response body spec

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/AdditionalInformation.kt
jldparker527: `addressId` is the correct name however, we don't need it as we just use the callback url in the detailUrl. We also no longer need to publish the `deliusId` in our address.created event

jldparker527:
Preliminary work is needed
DB changes
- https://dsdmoj.atlassian.net/browse/CPR-1146
- https://dsdmoj.atlassian.net/browse/CPR-1172
- https://dsdmoj.atlassian.net/browse/CPR-1176

Change existing behaviour 
- https://dsdmoj.atlassian.net/browse/CPR-1148

Migrations
- https://dsdmoj.atlassian.net/browse/CPR-1147

jldparker527:
The was done before the change in how we consume Delius events as a whole. All the the changes in the PR (except publishing a CPR event) are now done

PR #1990, opened 2026-04-27
CPR-1143 Add probation GET address endpoint

lbevan:


lbevan:
Do we need to consider merged records here? We can't really return another address, so would we return a 404? Not sure.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
```diff
@@ -2,8 +2,11 @@ package uk.gov.justice.digital.hmpps.personrecord.api.model.canonical
 
 import io.swagger.v3.oas.annotations.media.Schema
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.AddressEntity
+import java.util.UUID
 
 data class CanonicalAddress(
+  @Schema(description = "CPR address Id", example = "ec4c7479-218c-4f11-a02d-edd749820679")
+  val cprAddressId: UUID,
```
JunaedAhmedMOJ: should this not be String if we are adding this to the Canonical Api? Similar to cprUUID

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/AddressRepository.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.AddressEntity
+import java.util.UUID
+
+@Repository
+interface AddressRepository : JpaRepository<AddressEntity, Long> {
+
+  fun findByUpdateIdAndPersonCrn(updateId: UUID, personCrn: String): AddressEntity?
```
jldparker527, line 11: Do we want to consider indexing `update_id` column as this endpoint is expected to be called ofter

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
lbevan: I'm not sure of the best practices here. We could use a String, but UUID gives us type safety and the value is correctly parsed to JSON as is.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/AddressRepository.kt
lbevan: Probably a good idea to do this yeah, given the additional requests we're going to get via this route. Will add a ticket to the backlog.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
JunaedAhmedMOJ: generally with others APIs and this one we are using strings for IDs. So i guess we should try to continue that practice here

lbevan: Ok, makes sense to be consistent so have updated to String.

JunaedAhmedMOJ:
> Do we need to consider merged records here? We can't really return another address, so would we return a 404? Not sure.

worth asking Tony about this

JunaedAhmedMOJ (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/AddressRepository.kt
lbevan: https://dsdmoj.atlassian.net/browse/CPR-1162

jldparker527 (approved):

PR #1991, opened 2026-04-27
CPR-1140 renamed 'PersonDomainEventPublisher' to 'DomainEventPublisher'

jldparker527:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/DomainEvent.kt
```diff
@@ -9,6 +9,10 @@ data class DomainEvent @JsonCreator constructor(
   @JsonProperty("eventType") val eventType: String,
   @JsonProperty("personReference") val personReference: PersonReference? = null,
   @JsonProperty("additionalInformation") val additionalInformation: AdditionalInformation? = null,
```
jldparker527, line 11: Kept as AdditionalInformation for now instead of keeping Map<String, String>.

AdditonalInfomration is heavily wired into all the test.

Can potentially do it in another PR but I can't see any immediate reason to

jldparker527:
Do we wanna make place all our future domain events inside `PersonDomainEventListener`? if so could rename this to `DomainEventListener`

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/DomainEvent.kt
```diff
@@ -9,6 +9,10 @@ data class DomainEvent @JsonCreator constructor(
...
+  @JsonProperty("version") val version: Int? = null,
```
lbevan: This should probably default to 1 so that published events have a value by default.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/DomainEvent.kt
```diff
@@ -9,6 +9,10 @@ data class DomainEvent @JsonCreator constructor(
...
+  @JsonProperty("description") val description: String? = null,
+  @JsonProperty("detailUrl") val detailUrl: String? = null,
+  @JsonProperty("occurredAt") val occurredAt: String? = null,
```
lbevan: I think `version` and `occurredAt` are mandatory in the schema so they can be non nulls.
https://dsdmoj.atlassian.net/wiki/spaces/NDSS/pages/3732570166/ADR+-+Domain+Event+Schema

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/DomainEvent.kt
lbevan: Would be nice to do this at somepoint because having fields that only apply to some requests is confusing. But yeah, happy for that to be a later separate PR.

lbevan:
> Do we wanna make place all our future domain events inside `PersonDomainEventListener`? if so could rename this to `DomainEventListener`

Good question. Not sure right now but we could separate based on entity, i.e `PersonDomainEventListener` for person level events, `AddressDomainEventListener` for address level events. We may end up having others too, in which case 1 class would get quite large.
Can be approached as/when we hit it.

jldparker527:
> > Do we wanna make place all our future domain events inside `PersonDomainEventListener`? if so could rename this to `DomainEventListener`
> 
> Good question. Not sure right now but we could separate based on entity, i.e `PersonDomainEventListener` for person level events, `AddressDomainEventListener` for address level events. We may end up having others too, in which case 1 class would get quite large. Can be approached as/when we hit it.

agree

lbevan (approved):
👍

PR #1993, opened 2026-04-28
CPR-1144 Add probation POST address endpoint

lbevan:


lbevan:
Happy to split some of the service work out, though it will need to go in before this. LMK.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/probation/Address.kt
```diff
@@ -0,0 +1,44 @@
+package uk.gov.justice.digital.hmpps.personrecord.api.model.probation
+
+import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.personrecord.model.types.AddressStatusCode
+import uk.gov.justice.digital.hmpps.personrecord.model.types.CountryCode
+import java.time.LocalDate
+
+@Schema(name = "ProbationCreateAddress")
+data class Address(
+  @Schema(description = "Is the person without a permanent residence", example = "false", required = true)
```
lbevan, line 10: Using `Address` as class name to be consistent, but should we consider just using explicit naming so we dont have to alias in other files?
In this case, to avoid swagger name clashes we need to provide a specific schema name.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/person/Address.kt
```diff
@@ -120,6 +122,26 @@ data class Address(
...
+      buildingNumber = address.buildingNumber,
+      thoroughfareName = address.thoroughfareName,
+      dependentLocality = address.dependentLocality,
+      postTown = address.postTown,
+      county = address.county,
+      countryCode = address.countryCode,
+      comment = address.comment,
+      statusCode = address.statusCode,
+      usages = address.usages.map { AddressUsage.from(it) },
+      contacts = address.contacts.mapNotNull { Contact.from(it) },
```
jldparker527, line 142: Add deliusAddressId and isVerified here?

jldparker527: ...once branch is brought up to date

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAddressPostAPIControllerTest.kt
```diff
@@ -0,0 +1,230 @@
...
+    @Test
+    fun `should return 404 not found when probation record does not exist`() {
+      sendPostRequestAsserted<Unit>(
+        url = probationAddressApiUrl(randomCrn()),
+        body = createRandomAddress(),
+        roles = listOf(PROBATION_API_READ_WRITE),
+        expectedStatus = HttpStatus.NOT_FOUND,
+      )
+    }
+
```
jldparker527, line 98: Once brought up to date... test that asserts an address is not created when one with deliusAddressId already exists

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAddressPostAPIControllerTest.kt
```diff
@@ -0,0 +1,230 @@
...
+import uk.gov.justice.digital.hmpps.personrecord.test.randomContactType
+import uk.gov.justice.digital.hmpps.personrecord.test.randomCountryCode
+import uk.gov.justice.digital.hmpps.personrecord.test.randomCrn
+import uk.gov.justice.digital.hmpps.personrecord.test.randomName
+import uk.gov.justice.digital.hmpps.personrecord.test.randomPhoneNumber
+import uk.gov.justice.digital.hmpps.personrecord.test.randomPostcode
+import uk.gov.justice.digital.hmpps.personrecord.test.randomUprn
+import uk.gov.justice.digital.hmpps.personrecord.test.randomZonedDateTime
+
+class ProbationAddressPostAPIControllerTest : WebTestBase() {
```
jldparker527, line 29: Have a Probation address test base class here https://github.com/ministryofjustice/hmpps-person-record/pull/2051/changes#diff-9c141e3e8ff096180b883f6d6566b4a34c7287e9c96f501f9b50bddede7274e4

Could do with us working this in by either include it in this PR or reworking it later 🙃

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/person/Address.kt
lbevan: This is the `from` for the create address model. No requirement to support `isVerified` or `deliusAddressId` fields for this POST api.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/address/AddressService.kt
```diff
@@ -0,0 +1,77 @@
...
+  private fun create(address: Address, personEntity: PersonEntity): AddressEntity {
+    val matchingFieldsBeforeUpdate = PersonMatchRecord.from(personEntity)
+    val addressToSave = AddressEntity.from(address)
+    addressToSave.person = personEntity
+    personEntity.addresses.add(addressToSave)
+
+    val addressEntity = addressRepository.save(addressToSave)
+
+    reclusterIfMatchingFieldsChangedAndNotPassiveRecord(personEntity, matchingFieldsBeforeUpdate)
+
```
jldparker527, line 48: Publishing of any events?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/address/AddressService.kt
```diff
@@ -0,0 +1,77 @@
...
+    return addressEntity
+  }
+
+  private fun reclusterIfMatchingFieldsChangedAndNotPassiveRecord(
+    personEntity: PersonEntity,
+    matchingFieldsBeforeUpdate: PersonMatchRecord,
+  ) {
+    val matchingFieldsChanged = matchingFieldsBeforeUpdate.matchingFieldsAreDifferent(personEntity)
+    if (matchingFieldsChanged && personEntity.isNotPassive()) {
```
jldparker527, line 67: PersonService also has a `person.behaviour.reclusterOnUpdate` condition. Do we want this here to?

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAddressPostAPIControllerTest.kt
lbevan: The POST endpoint doesn't accept `deliusAddressId`, so it wouldn't be a check specifically for this, but agree that we should add a constraint and test it at somepoint.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAddressPostAPIControllerTest.kt
lbevan: This is an endpoint test class which needs a lot of functions from `WebTestBase` and `WebTestSetup`. Not sure its worth doing that with this test class. Though it does raise a good point, we should consider writing an E2E/larger probation integration test class to do create -> update -> delete flows.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/person/Address.kt
lbevan: May need to add `typeVerified`.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/address/AddressService.kt
lbevan: Good shout, but it looks like `person.behaviour.reclusterOnUpdate` is only on the Person DTO model. We only use the entity here so there is no need to worry about it for now.
There may be a need to think about this, or something similar, when handling updates via merge events.

jldparker527 (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/person/Address.kt
lbevan: `typeVerified` to be added under CPR-1190.

PR #2000, opened 2026-04-29
CPR-1143 Add probation GET address endpoint

lbevan:


lbevan:
Re-added this code that was previously reverted.
The country field is now maintained as a String and countryCode has been added separately.
Interesting to note that even though the field was on the CanonicalAddress, we never previously returned data for it.
Worth thinking about we handle this type of change going forward.

markreesmoj (approved):
👍

PR #2001, opened 2026-04-29
CPR-1161 permissions

markreesmoj:
Intention is to fix these alerts
https://github.com/ministryofjustice/hmpps-person-record/security/code-scanning/54
https://github.com/ministryofjustice/hmpps-person-record/security/code-scanning/55

Also pinning the github action to the hash, not just the version


lbevan (approved):
👍

PR #2002, opened 2026-04-29
CPR-1161-fix-workflow-security

JunaedAhmedMOJ:
fixes issue of Log Injection in securities.
Do not need to log as we do not get enough unknowns

We have only seen 3 of these Unknowns appearing in logs since January (all of which were in February), therefore we can delete this logging information


lbevan (approved):
👍

PR #2030, opened 2026-05-07
CPR-1181 Map Probation address usage

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/offender/ProbationAddress.kt
```diff
@@ -22,4 +24,12 @@ data class ProbationAddress(
...
   @JsonProperty("typeVerified")
   val isVerified: Boolean? = null,
+  @JsonProperty("type")
+  val usage: ProbationAddressUsage? = null,
+)
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class ProbationAddressUsage(
+  val code: String,
+  val description: String,
```
jldparker527, line 34: may need to be nullable

lbevan: The inner fields are marked as mandatory on the API 🤔 I'd imagine the top level usage fields would be null if there is no usage set.

JunaedAhmedMOJ:
Looks good , will need some testing on Dev to make sure

JunaedAhmedMOJ (approved):

PR #2033, opened 2026-05-07
CPR-1181 Modify address date field types to timestamp

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


lbevan:
Changing the type of these fields to timestamp to support the new Probation address start/end datetime fields.
Postgres correctly changes the type and maintains the date, setting the time to the default of `00:00:00.000` (see attached). While the default time will remain for Prison and Court records, we _should_ get the correct times when we migrate the Probation records.
Screenshot 2026-05-07 at 15 32 09

jldparker527 (approved):

PR #2034, opened 2026-05-07
CPR-1181 Modify Probation address start/end date to be LocalDateTime

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [x] 🔨 Have any downstream breaking changes
* [ ] 🚩 Need Feature flagged


lbevan:
Branched out of #2033. Potential option for switching to use LocalDateTime for these fields.
One downside in this PR is `ApiResponseSetupAddress` has its start/end date field changed to `LocalDateTime`, which is not accurate for all API responses.

jldparker527:
Outstanding question: Is it ok that ALL existing API calls will not return a DateTime format for address when retrieving a person?

src/main/resources/db/migration/V197__modify_address_date_time_field_types.sql
```diff
@@ -0,0 +1,11 @@
+BEGIN;
```
jldparker527, line 1: Seperate PR here https://github.com/ministryofjustice/hmpps-person-record/pull/2033

That to be merged first if this is to go in

lbevan: Closed #2033 because this needs to go in all at once, otherwise the canonical endpoints will break downstream consumers by returning datetime.

JunaedAhmedMOJ (dismissed):

src/main/resources/db/migration/V197__modify_address_date_time_field_types.sql
lbevan: Image

jldparker527 (approved):

PR #2048, opened 2026-05-13
CPR-1181 added delius id and type verified to address entity mapping

jldparker527:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


jldparker527:
reads `deliusAddressId` and `isVerified` from the database and proves when updating a probation person (without updating addresses) that it does not re-create address records

lbevan (approved):
👍

PR #2059, opened 2026-05-19
CPR-1195 Publish internal address events

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


lbevan:
Opted to flatten the event structure for these as there is no need for an intermediary like `PersonEventListener`. These address events can then be used by the external event publisher.

We also agreed to continue using `CPR_RECORD_CREATED` and `CPR_RECORD_UPDATED` instead of new address specific ones.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/listeners/EventLogListener.kt
```diff
@@ -18,4 +21,20 @@ class EventLogListener(
   fun onEventLogEvent(eventLog: RecordEventLog) {
     eventLogService.logEvent(eventLog)
   }
+
+  @Async
+  @TransactionalEventListener
+  fun onAddressCreated(addressCreated: AddressCreated) {
+    if (addressCreated.matchingFieldsHaveChanged) {
+      eventLogService.logEvent(RecordEventLog(CPR_RECORD_UPDATED, addressCreated.addressEntity.person!!))
```
markreesmoj, line 29: I know it would look odd but it seems that these events actually need a PersonEntity rather than an AddressEntity

markreesmoj: we could just publish a PersonUpdated event in both cases...

lbevan: I guess thats somewhat true, though I was trying to preempt some of the external events work that will be working with an Address. From a `what just happened` perspective, its an Address that was updated, its just at the moment the existing listeners just care about the person (based on what we agreed an hour ago).

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/address/AddressService.kt
```diff
@@ -60,6 +66,8 @@ class AddressService(
     val matchingFieldsChanged = matchingFieldsBeforeUpdate.matchingFieldsAreDifferent(addressEntity.person!!)
     tryRecluster(addressEntity.person!!, matchingFieldsChanged)
```
markreesmoj, line 67: nitpicking again but it seems we only want to recluster and publish this event if the matching fields have changed

We could do
```
if(matchingFieldsBeforeUpdate.matchingFieldsAreDifferent(addressEntity.person!!)){
tryRecluster(addressEntity.person!!)
publisher.publishEvent(AddressUpdated(addressEntity)
}
```

which would save a couple of if statements and arguments being passed around

lbevan: Good shout. Will take a look at this.

lbevan: Actually, I'm not we'd want conditionalise the `publishEvent` here. Consumers are making the determination whether to act on an event like this; telemetry will always fire, event log will only be when matching fields change. This matches the pattern for `PersonService`.

As for the checks around recluster, 2 places (create and update) need to do the same checks and recluster, and the event publishing also needs the `matchingFieldsChanged `. I can move the `if` checks for `passive` and `matchingFieldsChanged` out and rename `tryRecluster` -> `recluster`, but we'd be spreading the checks everywhere :/.

markreesmoj (approved):
👍

PR #2062, opened 2026-05-19
CPR-1148 added address service delete function

jldparker527:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/address/AddressService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/address/AddressService.kt
@@ -73,10 +73,7 @@ class AddressService(
 
     addressEntity.person = null
     personRepository.save(personEntity)
-    if (personEntity.isNotPassive()) {
-      personMatchService.saveToPersonMatch(personEntity)
-      personEntity.personKey?.let { reclusterService.recluster(personEntity) }
-    }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/address/AddressService.kt
```diff
@@ -63,6 +65,20 @@ class AddressService(
...
 
+  @Transactional
+  fun deleteAddress(addressEntity: AddressEntity) {
+    val personEntity = addressEntity.person!!
+    val doesExist = personEntity.addresses.remove(addressEntity)
+    if (!doesExist) return
+
+    addressEntity.person = null
+    personRepository.save(personEntity)
+    if (personEntity.isNotPassive()) {
```
markreesmoj: Is there a test for this `if` condition?

markreesmoj: could call `tryRecluster` with `matchingFieldsChanged` true instead of writing a test?

jldparker527: 👍

markreesmoj (approved):
👍

PR #2064, opened 2026-05-20
CPR-1148 removed migration script 199 and 200

jldparker527:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/resources/db/migration/V199__add_unique_constraint_to_address_delius_address_id.sql
+++ b/src/main/resources/db/migration/V199__add_unique_constraint_to_address_delius_address_id.sql
@@ -1,8 +1,0 @@
-BEGIN;
--------------------------------------------------------
-
-ALTER TABLE IF EXISTS address
-    ADD CONSTRAINT unique_delius_address_id UNIQUE (delius_address_id);
-
------------------------------------------------------
-COMMIT;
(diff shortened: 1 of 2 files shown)


jldparker527:
Screenshot 2026-05-20 at 11 02 34

removed 199 and 200 from DEV database

jldparker527:
Context:
Originally deployed this to DEV https://github.com/ministryofjustice/hmpps-person-record/pull/2061/changes

Tried to deploy to PRE_PROD but the migration appeared to be taking to long and more work is needed to accomidate this migration (Still needs to be done and is actually good it didn't get deployed all the way up... see below)

Then we found we was getting unique constraint errors in DEV the morning after (investigation still ongoing as to why)
So to unblock any deployments in the interim, we reverted and deployed this to DEV https://github.com/ministryofjustice/hmpps-person-record/pull/2063/changes

However, I forget that we originally had issues deploying to pre PROD so this would still block deployments

So finally, we need to either:
1. manually delete the two flyway records from DEV db and delete the 199 and 200 script from main
2. finish the work in actually allowing preprod to wait for long running migration script

Have gone with option 1 in favor of a fast fix, which is this change

markreesmoj (approved):
199 has not been applied in preprod so 👍

PR #2065, opened 2026-05-20
CPR-1148 initial fix for allowing the adding of unique constraint to …

jldparker527:
…delius address id

# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged
* [x] 🔀 Require any database migrations
* [x] ☁️ Need to provision/update any cloud platform resources

### DIFF
--- a/helm_deploy/hmpps-person-record/values.yaml
+++ b/helm_deploy/hmpps-person-record/values.yaml
@@ -28,7 +28,7 @@ generic-service:
       path: /health/liveness
       port: http
     initialDelaySeconds: 30
-    periodSeconds: 10
+    periodSeconds: 60
     timeoutSeconds: 10
     failureThreshold: 10
 
@@ -37,7 +37,7 @@ generic-service:
       path: /health/readiness
       port: http
     initialDelaySeconds: 30
-    periodSeconds: 10
+    periodSeconds: 60
     timeoutSeconds: 10
     failureThreshold: 10
 
--- a/helm_deploy/hmpps-person-record/values.yaml
+++ b/helm_deploy/hmpps-person-record/values.yaml
@@ -23,6 +23,24 @@ generic-service:
     tag: app_version # override at deployment time
     port: 8080
 
+  livenessProbe:
+    httpGet:
+      path: /health/liveness
+      port: http
+    initialDelaySeconds: 30
+    periodSeconds: 10
+    timeoutSeconds: 10
+    failureThreshold: 10
+
+  readinessProbe:
+    httpGet:
+      path: /health/readiness
+      port: http
+    initialDelaySeconds: 30
+    periodSeconds: 10
+    timeoutSeconds: 10
+    failureThreshold: 10
+
   ingress:
     enabled: true
     host: app-hostname.local # override per environment
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/extensions/PersonEntityExtensions.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/extensions/PersonEntityExtensions.kt
@@ -1,7 +1,9 @@
 package uk.gov.justice.digital.hmpps.personrecord.extensions
 
+import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.AddressEntity
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.ContactEntity
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.ReferenceEntity
+import uk.gov.justice.digital.hmpps.personrecord.model.person.Address
 import uk.gov.justice.digital.hmpps.personrecord.model.types.ContactType
 import uk.gov.justice.digital.hmpps.personrecord.model.types.IdentifierType
 import uk.gov.justice.digital.hmpps.personrecord.model.types.IdentifierType.CRO
@@ -24,7 +26,7 @@ fun <T, E> T.existsIn(childEntities: List<E>, match: (T, E) -> Boolean, yes: (E)
   return found?.let { yes(found) } ?: no()
 }
 
-fun <T, E> T.existsIn(childEntities: List<E>, match: (T, E) -> Boolean, yes: (T, E) -> E, no: () -> E?): E? {
+fun Address.existsIn(childEntities: List<AddressEntity>, match: (Address, AddressEntity) -> Boolean, yes: (Address, AddressEntity) -> AddressEntity, no: () -> AddressEntity?): AddressEntity? {
   val found = childEntities.find { match(this, it) }
   return found?.let { yes(this, found) } ?: no()
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/extensions/PersonEntityExtensions.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/extensions/PersonEntityExtensions.kt
@@ -23,3 +23,8 @@ fun <T, E> T.existsIn(childEntities: List<E>, match: (T, E) -> Boolean, yes: (E)
   val found = childEntities.find { match(this, it) }
   return found?.let { yes(found) } ?: no()
 }
+
+fun <T, E> T.existsIn(childEntities: List<E>, match: (T, E) -> Boolean, yes: (T, E) -> E, no: () -> E?): E? {
+  val found = childEntities.find { match(this, it) }
+  return found?.let { yes(this, found) } ?: no()
+}
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/AddressEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/AddressEntity.kt
@@ -119,7 +119,7 @@ class AddressEntity(
   var version: Int = 0,
 ) {
 
-  fun update(address: Address): AddressEntity {
+  fun update(address: Address) {
     this.noFixedAbode = address.noFixedAbode
     this.startDate = address.startDate
     this.endDate = address.endDate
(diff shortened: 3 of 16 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/extensions/PersonEntityExtensions.kt
```diff
@@ -23,3 +23,8 @@ fun <T, E> T.existsIn(childEntities: List<E>, match: (T, E) -> Boolean, yes: (E)
   val found = childEntities.find { match(this, it) }
   return found?.let { yes(found) } ?: no()
 }
+
```
jldparker527, line 28: Simply just created an additional method here as it's used by all other child class types (contacts, sentences, etc...) to limit surface area. If something goes horrible wrong, it will only effect address and not all other child classes

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/builder/AddressBuilder.kt
```diff
@@ -12,10 +12,19 @@ object AddressBuilder {
     address.existsIn(
       childEntities = personEntity.addresses,
       match = { ref, entity -> ref.matches(entity) },
-      yes = { it },
+      yes = { ref, entity -> entity.update(ref) },
```
jldparker527: what we are saying is:
GIVEN the address record already exists either by
- all the fields match
- or some fields don't match but have the same deliusAddressId as whats in the DB
THEN update the fields


Because we have temporarily updated the matches function to return true if fields don't match but deliusAddressId matches, we now need to update any fields also on existing address record

jldparker527:
Context...

We attempted to merge in just the migration file change previously but this caused constraint errors in dev. 

This happened because when we consumed probation person events with updated addresses in them, we treated them as new records and attempted to INSERT them. But the actual record they sent us have the same deliusAddressId as ones we already have saved in our DB.


I believe we can revert all the other changes in the PR except the migration script once we have deployed the final work around how we consume probation address

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/builder/AddressBuilder.kt
```diff
@@ -12,10 +12,19 @@ object AddressBuilder {
...
       no = { AddressEntity.from(address) },
     )
   }
 
-  private fun Address.matches(entity: AddressEntity): Boolean = this == Address.from(entity)
+  // Once the final work around changing how we consume probation address
+  // is done, this can be reverted back
+  private fun Address.matches(entity: AddressEntity): Boolean = when {
+    this == Address.from(entity) -> true
+    entity.deliusAddressId == this.deliusAddressId -> when {
```
jldparker527: most address will have a null deliusAddressId so this doesn't prove equality.

once we have asserted both fields are not null, then check them for equality

lbevan:
What is the plan for existing duplicate ids?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/AddressEntity.kt
```diff
@@ -140,6 +140,7 @@ class AddressEntity(
     this.isVerified = address.isVerified
     this.recordType = address.recordType
     updateChildEntities(address)
+    return this
```
markreesmoj: can we avoid this return? It will be modified anyway

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/builder/AddressBuilder.kt
```diff
@@ -12,10 +12,19 @@ object AddressBuilder {
...
+      entity.deliusAddressId == null && this.deliusAddressId == null -> false
```
markreesmoj: this is run for everything or just probation?

I know it's a bit old-fashioned but would this do the trick?
```
if(this.deliusAddressId != null){
  return entity.deliusAddressId == this.deliusAddressId
}
return this == Address.from(entity)
```

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/builder/AddressBuilder.kt
markreesmoj: good explanation - I think I follow it

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/extensions/PersonEntityExtensions.kt
markreesmoj: how about we make it non-generic then?
`Address.existsIn` would be a lot easier to understand 

Took me ages to figure out that the difference is the method signature of the `yes` function (at least I think that is the difference)

markreesmoj:
> What is the plan for existing duplicate ids?

I don't think there are any? You mean deliusAddressIDs? Hopefully there aren't any - there weren't in dev. We can run a query to check easily enough

markreesmoj:
might as well increase the readiness/liveness timeout in this PR as we will need it to go to preprod & prod if successful

lbevan:
> > What is the plan for existing duplicate ids?
> 
> I don't think there are any? You mean deliusAddressIDs? Hopefully there aren't any - there weren't in dev. We can run a query to check easily enough

There are 2 in Preprod and 2 in Prod. I think we'd have to choose 1 of them for each case and delete the other manually before progressing the script. Likely we can take the latest primary id.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/extensions/PersonEntityExtensions.kt
jldparker527: yes that would be nicer. Uust did a straight copy as this change want have to stay around for long

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/AddressEntity.kt
jldparker527: No becuase we need it to return in the `yes` flow

`yes = { ref, entity -> entity.update(ref) }` <- needs the entity returned

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/builder/AddressBuilder.kt
jldparker527: yes much nicer 👍

jldparker527:
Action plan

DEV
Check still no duplicates
Deploy fast
Test

PRE PROD
Gather duplicates and deletes oldest
Deploy fast
Test

PROD
Gather duplicates and deletes oldest
Deploy fast
Validate

helm_deploy/hmpps-person-record/values.yaml
```diff
@@ -23,6 +23,24 @@ generic-service:
     tag: app_version # override at deployment time
     port: 8080
 
+  livenessProbe:
```
jldparker527, line 26: I think changes here are applied to all envs?

jldparker527: @markreesmoj

helm_deploy/hmpps-person-record/values.yaml
```diff
@@ -23,6 +23,24 @@ generic-service:
...
+
+  readinessProbe:
+    httpGet:
+      path: /health/readiness
+      port: http
+    initialDelaySeconds: 30
+    periodSeconds: 10
+    timeoutSeconds: 10
+    failureThreshold: 10
+
```
markreesmoj, line 43: these are the existing settings - we just need to override the ones we want to. 


```suggestion
    periodSeconds: 60
```
should give us 30 to start with then 10 * 60 seconds so ten and a half minutes
Hopefully enough to apply a constraint. If not we can tweak

As I am unsure whether readiness or liveness is the one in play let's just change both (unless you want to read the docs)

jldparker527: 👍

helm_deploy/hmpps-person-record/values.yaml
markreesmoj: won't make any odds in dev as it was fast enough there anyway.
Wondering if the problem in preprod was actually that the constraint could not be applied because of the duplicates?
either way can just up it as it will only allow more time, will not take much longer

helm_deploy/hmpps-person-record/values.yaml
```diff
@@ -28,7 +28,7 @@ generic-service:
       path: /health/liveness
       port: http
     initialDelaySeconds: 30
-    periodSeconds: 10
+    periodSeconds: 60
     timeoutSeconds: 10
     failureThreshold: 10
```
markreesmoj, line 34:
```

livenessProbe:
  periodSeconds: 60

```

Is all that needs to be set - the other values will be defaulted to those specified here which are from the generic service.

Doesn't matter for this PR, just making it clear how this file works

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/extensions/PersonEntityExtensions.kt
```diff
@@ -23,3 +25,8 @@ fun <T, E> T.existsIn(childEntities: List<E>, match: (T, E) -> Boolean, yes: (E)
...
+fun Address.existsIn(childEntities: List<AddressEntity>, match: (Address, AddressEntity) -> Boolean, yes: (Address, AddressEntity) -> AddressEntity, no: () -> AddressEntity?): AddressEntity? {
```
markreesmoj, line 29: 👍

markreesmoj (approved):
👍

JunaedAhmedMOJ (approved):

PR #2051, opened 2026-05-14
CPR-1148 Splits out probation events from main flow

jldparker527:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🔨 Have any downstream breaking changes
* [ ] 🚩 Need Feature flagged
* [ ] 🔀 Require any database migrations
* [ ] 🔔 Alerting of any other teams of changes
* [ ] ☁️ Need to provision/update any cloud platform resources
* [ ] ⚡️ Benefit from running the [load tests](https://github.com/ministryofjustice/hmpps-person-record-gatling)

PR #2066, opened 2026-05-20
CPR-1141 address id migration probation

markreesmoj:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged
* [x] 🔔 Alerting of any other teams of changes


jldparker527:
This to only be triggered after the following are deployed in all envs

1. Unique constraint https://github.com/ministryofjustice/hmpps-person-record/pull/2065
2. Delete to AddressService https://github.com/ministryofjustice/hmpps-person-record/pull/2062
3. Change how we handle probation address events https://github.com/ministryofjustice/hmpps-person-record/pull/2051
4. Optional: revert the change in PR one (except the migration script)

lbevan:
If we do this before CPR-1148 then it looks good.
If we do this after CPR-1148 then we will have to update the addresses separately.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/seeding/TransactionalProbationUpdater.kt
```diff
@@ -0,0 +1,39 @@
...
+  fun update(case: ProbationCase) {
+    val person = Person.from(case)
+    personRepository.findByCrn(person.crn!!).exists(
+      no = {
+        log.error("CRN not found in Database ${person.crn}")
+      },
+      yes = {
+        if (it.isNotMerged()) {
+          it.update(person)
+          personRepository.save(it)
```
jldparker527, line 26: Do we want to publish any internal events?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/seeding/PopulateFromProbationController.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.hibernate.query.Page.page
+import org.slf4j.LoggerFactory
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.personrecord.client.CorePersonRecordAndDeliusClient
+import uk.gov.justice.digital.hmpps.personrecord.client.CorePersonRecordAndDeliusClientPageParams
+
+@RestController
+class PopulateFromProbationController(
```
jldparker527, line 16: As it stands, don't see how running 6 pods at the same time would work as each pod would only now the start page but not the end page. Looks like we would just have 6 pods running and redoing work

lbevan: If we wanted to run on 1 pod, we could utilise atomics, `SupervisorJob() + Dispatchers.IO` scopes and fixed (configurable) number of workers. Bit more complicated to setup codewise and potential resource concerns.

markreesmoj: one pod starts on page 0, one on 1000 etc etc. Once they have all done 1000 pages (or once all the addresses have deliusAddressIds) kill them
crude but simple :-)

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/seeding/TransactionalProbationUpdater.kt
markreesmoj: should not be any changes to existing matching data so I don't think so

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/seeding/PopulateFromProbationController.kt
markreesmoj: we could supply a start and end I suppose but we really need to get going on this. Maybe one to look at today if the above is not fast enough

jldparker527: Do we now how pages will be returned in prod?

jldparker527: Thinks it's worth asking them how many address they have in prod then we know what we are working with and wether or need to optimize it or keep it simple

markreesmoj: we know how many, we already have them

markreesmoj: Depends on page size. There are 2.2 million probation records so on page size of 1000 that is 2200 pages

markreesmoj: Preprod figures
2375746 probation records
7116660 probation addresses (so about 3 per person on average)
176000 address records in preprod have deliusAddressId populated

jldparker527 (approved):

PR #2067, opened 2026-05-21
CPR-1190 Add typeVerified to Probation create address endpoint

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAddressPostAPIControllerTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAddressPostAPIControllerTest.kt
@@ -195,6 +195,7 @@ class ProbationAddressPostAPIControllerTest : WebTestBase() {
     countryCode = randomCountryCode(),
     comment = randomName(),
     statusCode = randomAddressStatusCode(),
+    typeVerified = randomBoolean(),
     usages = listOf(AddressUsage(randomAddressUsageCode(), randomBoolean())),
     contacts = listOf(AddressContact(randomContactType(), randomPhoneNumber(), "44")),
   )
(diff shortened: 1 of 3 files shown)


lbevan:
Defaulting to `false` to not break SAS in the short term. We may be able remove the default once they've mapped it.

markreesmoj (dismissed):
👍

lbevan:
Just checking with SAS whether we need to default this to `true` in the short term as they are currently only send us confirmed addresses.

lbevan:
Will need to default to `true` in the short term because SAS are only sending confirmed addresses.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAddressPostAPIControllerTest.kt
```diff
@@ -195,6 +195,7 @@ class ProbationAddressPostAPIControllerTest : WebTestBase() {
     countryCode = randomCountryCode(),
     comment = randomName(),
     statusCode = randomAddressStatusCode(),
+    typeVerified = randomBoolean(),
```
markreesmoj: If we understand correctly, the required behaviour here is that typeVerified is always true 
SAS are not always (ever?) going to supply it so we need to default it
If so we should test that behaviour too
Code looks fine

lbevan: The `true` default is the short term while they haven't mapped it, Greg will map this over the next week or two at which point they will likely send `true` as they currently only expect to send `confirmed` addresses, though there are discussions for this to potentially change. Once their change is done we can make this field mandatory with no default.

I've changed the existing tests to send `true` and added an new test where `typeVerified` is not supplied.

markreesmoj: I think I get it now. Because we do not specify typeVerified in `createRandomAddress` it will default to true on the Address object
We remove it from the JSON sent and then the API defaults it to true (also using the  Address object)
The comparison is therefore always  true == true

markreesmoj (approved):
👍

PR #2073, opened 2026-05-26
CPR-1147 save new records

markreesmoj:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


lbevan (dismissed):
👍

jldparker527 (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/seeding/TransactionalProbationUpdater.kt
```diff
@@ -19,6 +25,8 @@ class TransactionalProbationUpdater(
     personRepository.findByCrn(person.crn!!).exists(
       no = {
         log.error("CRN not found in Database ${person.crn}")
+        personService.processPerson(person, { personRepository.findByCrn(person.crn) })
```
jldparker527, line 28: can't we just pass null into processPersons second argument

markreesmoj: `{ null }` would work fine - this call to the personRepository will return null anyway (unless we are on the last page and the record has just been created but we don't have it yet. Chances of that are just about zero though)

PR #2079, opened 2026-05-28
CPR-1150 Publish CPR address created domain event when SAS create an address

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [x] 🚩 Need Feature flagged


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/address/AddressService.kt
```diff
@@ -29,22 +30,23 @@ class AddressService(
     address: Address,
     findPerson: () -> PersonEntity?,
     findAddress: () -> AddressEntity?,
+    eventSource: DomainEventSource = DomainEventSource.CPR,
```
jldparker527: We may not want to default this now in case a call site in the future isn't forced to pass something in?

jldparker527 (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/address/AddressService.kt
lbevan: Removed default.

PR #2105, opened 2026-06-08
CPR-1192 Order Common Platform addresses by record type (Primary first)

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddressSorter.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddressSorter.kt
@@ -26,5 +26,3 @@ object CanonicalAddressSorter {
     null -> 2
   }
 }
-
-
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddressSorter.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddressSorter.kt
@@ -9,13 +9,9 @@ import uk.gov.justice.digital.hmpps.personrecord.model.types.SourceSystemType.CO
 
 object CanonicalAddressSorter {
 
-  private val orderingStrategies: Map<SourceSystemType, Comparator<AddressEntity>> = mapOf(
-    COMMON_PLATFORM to compareBy { it.recordType.recordTypeOrder() },
-  )
-
-  fun sort(addresses: List<AddressEntity>): List<AddressEntity> {
-    val comparator = addresses.sourceSystem()?.let { orderingStrategies[it] } ?: return addresses
-    return addresses.sortedWith(comparator)
+  fun sort(addresses: List<AddressEntity>): List<AddressEntity> = when (addresses.sourceSystem()) {
+    COMMON_PLATFORM -> addresses.sortedWith(compareBy { it.recordType.recordTypeOrder() })
+    else -> addresses
   }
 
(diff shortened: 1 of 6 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddressSorterTest.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    val sorted = CanonicalAddressSorter.sort(addressesFor(COMMON_PLATFORM, noRecordType, previous, primary))
+
+    assertThat(sorted).containsExactly(primary, previous, noRecordType)
+  }
+
+  @Test
+  fun `should keep original order for source systems without an ordering strategy`() {
+    listOf(NOMIS, DELIUS, LIBRA).forEach { sourceSystem ->
+      val previous = address(randomPostcode(), PREVIOUS)
```
markreesmoj: none of these source systems have PRIMARY or PREVIOUS addresses. I know this concept is already confusing because we made it up ourselves. This could lead someone to think that we do in fact have such addresses

lbevan: Clarified this test by setting `recordType` to null and updating naming.

lbevan:
We could consider merging `recordType` into `addressStatus`, where `Primary == Main`, which would provide the same functionality and then we could drop the `recordType` field. This is a bit more involved so if we want to do this we can do it separately, at least for now this fixes the bug.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddressSorter.kt
```diff
@@ -0,0 +1,28 @@
...
+object CanonicalAddressSorter {
+
+  private val orderingStrategies: Map<SourceSystemType, Comparator<AddressEntity>> = mapOf(
+    COMMON_PLATFORM to compareBy { it.recordType.recordTypeOrder() },
+  )
+
+  fun sort(addresses: List<AddressEntity>): List<AddressEntity> {
+    val comparator = addresses.sourceSystem()?.let { orderingStrategies[it] } ?: return addresses
+    return addresses.sortedWith(comparator)
+  }
```
markreesmoj:
```suggestion
  fun sort(addresses: List<AddressEntity>): List<AddressEntity> = when (addresses.sourceSystem()) {
    COMMON_PLATFORM -> addresses.sortedWith(compareBy { it.recordType.recordTypeOrder() })
    else -> addresses
  }
```

A small simplification to avoid the conditional return while preserving extensibility

markreesmoj:
> We could consider merging `recordType` into `addressStatus`, where `Primary == Main`, which would provide the same functionality and then we could drop the `recordType` field. This is a bit more involved so if we want to do this we can do it separately, at least for now this fixes the bug.

Maybe but we should gather requirements from anyone who wants this sorting before making a change. I would like to get rid of recordType if we can  - after all we made it up and none of our clients want or expect it.

markreesmoj (dismissed):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddressSorter.kt
lbevan: Very nice suggestion, applied :)

markreesmoj (approved):
👍

PR #2108, opened 2026-06-10
CPR-1203 alternative

markreesmoj:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/listeners/AddressDomainEventListener.kt
```diff
@@ -24,27 +25,34 @@ import java.time.format.DateTimeFormatter
...
 
   @TransactionalEventListener
   fun onAddressCreated(addressCreated: AddressCreated) {
-    publishAddressDomainEvent(
-      addressEntity = addressCreated.addressEntity,
-      eventSource = addressCreated.eventSource.identifier,
-      action = "created",
-      config = buildAddressDomainEventConfig(addressCreated.addressEntity.person!!.sourceSystem, CPR_PROBATION_ADDRESS_CREATED),
-    )
+    if (addressCreated.eventSource != DomainEventSource.CPR) {
```
lbevan, line 33: We use a CPR event source via the POST API too. We still need to publish in that scenario so we'd need to think about that.

PR #2113, opened 2026-06-10
CPR-1202 move addess events from delius which originated in CPR to a …

markreesmoj:
…different queue

Will need to merge https://github.com/ministryofjustice/cloud-platform-environments/pull/43180/changes and do the same in preprod and prod before releasing. The events which are being moved to another queue are only emitted in dev at the moment but the new queues need to exist otherwise the application will not start

# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged
* [x] ☁️ Need to provision/update any cloud platform resources


lbevan (commented):
Generally looks good.
Any circular test updates needed?

markreesmoj:
> Generally looks good. Any circular test updates needed?

For creates - don't think so - it publishes to the topic, the test passes and that should be enough. 
We could perhaps do with one for updates though - I will put one in - in fact the one from the other PR will do nicely

lbevan (approved):
👍

PR #2134, opened 2026-06-17
CPR-1202 ensure that all database interactions are within a transaction

markreesmoj:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationEventListener.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationEventListener.kt
@@ -49,7 +49,7 @@ class ProbationEventListener(
     addressService.processAddress(
       address = probationAddress,
       findPerson = { personRepository.findByCrn(event.getCrn())!! },
-      findAddress = { personRepository.findByCrn(event.getCrn())?.addresses?.firstOrNull { it.deliusAddressId == deliusAddressId } },
+      findAddress = { addressRepository.findByDeliusAddressId(deliusAddressId) },
       eventSource = DELIUS,
     )
   }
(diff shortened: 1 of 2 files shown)


lbevan (dismissed):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationEventListener.kt
```diff
@@ -45,32 +44,25 @@ class ProbationEventListener(
...
-    val personEntity = personRepository.findByCrn(crn)!!
-
+    val deliusAddressId = event.getDeliusAddressId()!!
+    val probationAddress = corePersonRecordAndDeliusClient.getAddress(deliusAddressId)!!
     addressService.processAddress(
       address = probationAddress,
-      findPerson = { personEntity },
-      findAddress = { personEntity.addresses.firstOrNull { it.deliusAddressId == probationAddress.deliusAddressId } },
+      findPerson = { personRepository.findByCrn(event.getCrn())!! },
+      findAddress = { personRepository.findByCrn(event.getCrn())?.addresses?.firstOrNull { it.deliusAddressId == deliusAddressId } },
```
markreesmoj: I wanted to replace this with `addressRepository.findByDeliusAddressId()` but a test failed. 
I am not sure there is really any value in checking whether the address belongs to the CRN from the event but this keeps the check in place

markreesmoj: We do not do this check for deletes.
In fact the more I think about it, the more I think we should ditch this check

JunaedAhmedMOJ (approved):

PR #2144, opened 2026-06-19
CPR-1152 find swagger classes with the same name

markreesmoj:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/SwaggerAvoidSameClassNameTest.kt
```diff
@@ -0,0 +1,22 @@
...
+  @Test
+  fun `ensure there are not two classes with the same name with Swagger annotations`() {
+    val kotlinFilesLocation = Paths.get("src/main/kotlin").toAbsolutePath()
+    val ls = File(kotlinFilesLocation.toString()).walkTopDown().filter {
+      it.isFile() && it.readText().contains("@Schema")
+    }.map { it.name }.groupingBy { it }.eachCount().filter { it.value > 1 }
+
+    if (ls.isNotEmpty()) {
+      ls.forEach { filename ->
+        print("There is more than one class named $filename with swagger annotations")
```
markreesmoj: output currently 
There is more than one class named AddressUsage.kt=3 with swagger annotationsThere is more than one class named Address.kt=2 with swagger annotations

need to figure out if this is actually a problem - nobody is complaining about it.

markreesmoj: This was the original inspiration: 
https://github.com/ministryofjustice/prison-api/blob/a805341ad44eaffd233092620805eff675fb3ffa/src/test/java/uk/gov/justice/hmpps/prison/api/resource/OpenApiDocsDuplicateTest.kt#L8

markreesmoj: and this describes the problem
https://github.com/swagger-api/swagger-core/issues/2861

lbevan: `AddressUsage.kt` - 1 for `ProbationCreateAddresss`, 1 for SysconSync, 1 for the core model. The core model doesn't look like its used for an endpoint so not sure why it has swagger annotations. The `ProbationCreateAddress` one is supplied a name so that is what it generates with instead of the default class name.

`Address.kt`- 1 for `ProbationCreateAddress` and 1 for SysconSync. The `ProbationCreateAddress` one is supplied a name so that is what it generates with instead of the default class name.

Maybe we should consider renaming these classes/files so that they are more descriptive for each source system rather than just `Address`. That way we can stop aliasing in files that use more than 1 and its easier to reason when searching.
I'm personally in favour of something like `ProbationAddressDto`, `ProbationCreateAddressDto` etc. Follows in the same vein with our entities.

lbevan (approved):
👍

PR #2146, opened 2026-06-22
SDIT-3925: 🗃️ Change not null fields and add indices

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/prison/PrisonReligion.kt
```diff
@@ -6,15 +6,15 @@ import java.time.LocalDate
 import java.time.LocalDateTime
 
 data class PrisonReligion(
-  val religionCode: String? = null,
-  val religionDescription: String? = null,
-  val changeReasonKnown: Boolean? = null,
+  val religionCode: String,
+  val religionDescription: String?,
```
petergphillips, line 10: this should really be not nullable too, but the current functionality returns `null` if the `code` isn't found when looking up the description

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/prison/PrisonReligionEntity.kt
```diff
@@ -35,25 +35,25 @@ class PrisonReligionEntity(
...
 
-  @Column(name = "start_date")
-  var startDate: LocalDate? = null,
+  @Column(name = "start_date", nullable = false)
+  var startDate: LocalDate,
 
   @Column(name = "end_date")
   var endDate: LocalDate? = null,
 
-  @Column(name = "modify_date_time", nullable = true)
```
petergphillips, line 53: default is `true` anyway so unnecessary

src/main/resources/db/migration/V200__modify_prison_religion_table.sql
```diff
@@ -0,0 +1,18 @@
...
+
+truncate table personrecordservice.prison_religion;
+
+ALTER TABLE IF EXISTS personrecordservice.prison_religion
+    ALTER COLUMN religion_code SET NOT NULL,
+    ALTER COLUMN start_date SET NOT NULL,
+    ALTER COLUMN change_reason_known SET NOT NULL;
+
+CREATE INDEX idx_prison_religion_prison_number
+    ON personrecordservice.prison_religion (prison_number);
```
petergphillips, line 10: added the two indices according to repository usage

src/main/resources/db/migration/V200__modify_prison_religion_table.sql
```diff
@@ -0,0 +1,18 @@
+BEGIN;
+-------------------------------------------------------
+
+truncate table personrecordservice.prison_religion;
```
petergphillips: we've only gone to dev so far with the migration so okay to truncate and start again

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/prison/PrisonReligion.kt
lbevan: We could create a `UNKNOWN("Unknown")` type as we've done with other enums.

IIRC, the reason we needed to store `code` as a string instead of an enum is because the value could be an old code we don't support? If that is not the case anymore we could consider doing that and drop the description.

petergphillips: Don't think that would work as all the old codes should be supported and descriptions should really always be set. Ideally the enum should be the full list of all possible NOMIS codes (107 in total).

petergphillips: I think either way, would perhaps be better to split out into a separate discussion as this change is non breaking

src/main/resources/db/migration/V200__modify_prison_religion_table.sql
jldparker527: We won't this in the migration script. Will want to manually run this in dev

petergphillips: I copied this from `V189__modify_prison_religion_table.sql` by @markreesmoj that did exactly the same thing. Don't mind though, happy whichever way is good.

markreesmoj: `select * from personrecordservice.prison_religion pr where pr.start_date is null or pr.change_reason_known is null or pr.religion_code is null`

returns 3 records in dev which look like they were done using manual testing in October
none in prod.
So the truncate is not really required to apply these changes - we could just delete those 3 records in dev

If we want to truncate before migrating in dev that is fine but we should probably do it manually as part of the migration process

lbevan (approved):
👍

PR #2147, opened 2026-06-22
CPR-1162-index-update-id

JunaedAhmedMOJ:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


src/main/resources/db/migration/V200__add_index_address_update_id.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
+-------------------------------------------------------
+CREATE INDEX idx_update_id ON address(update_id);
```
lbevan: Should this be a unique index?

jldparker527: Also as mentioned yesterday, we may want a composite index too for `findByUpdateIdAndPersonCrn`

markreesmoj (dismissed):
👍

markreesmoj:
Why are we actually doing this change though?
We only call this twice - one on SAS address deletes which is hardly going to be done frequently and the performance is unimportant, and once when we add the Delius address Id to the address, which is going to be short lived.
Seems pointless to add an index to speed those two up, especially since we do not have any known and measured performance problems there. 
I recommend not doing this change

lbevan:
Holding off on this for now. Not gaining much by adding it. Will see perf in prod.

PR #2148, opened 2026-06-22
CPR-1159 Refactor publishing of Person domain events

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/listeners/PersonDomainEventListener.kt
```diff
@@ -1,70 +1,20 @@
...
-import java.time.format.DateTimeFormatter
+import uk.gov.justice.digital.hmpps.personrecord.service.cprdomainevents.publishers.PersonEventPublisher
 
 @Profile("!preprod & !prod")
 @Component
-class PersonDomainEventListener(
-  private val domainEventPublisher: DomainEventPublisher,
-  @Value("\${core-person-record.base-url}") private val baseUrl: String,
-) {
+class PersonDomainEventListener(personEventPublishers: List<PersonEventPublisher>) {
```
lbevan, line 11: Need to keep this naming as it conflicts with another in the same package. We can consider collapsing once the feature flag is removed.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/publishers/PersonEventPublisher.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.personrecord.service.cprdomainevents.publishers
+
+import uk.gov.justice.digital.hmpps.personrecord.model.types.SourceSystemType
+import uk.gov.justice.digital.hmpps.personrecord.service.cprdomainevents.events.person.PersonCreated
+
+interface PersonEventPublisher {
```
jldparker527, line 6: Maybe be worth making this an abstract class so we can box the private method in all the publishers into it

jldparker527: something like this
https://github.com/ministryofjustice/hmpps-person-record/compare/CPR-1159-refactor-josh?expand=1

We could even remove onCreate() 🙃

markreesmoj: in Kotlin you can implement methods in interfaces, which is a bit wild

markreesmoj: https://kotlinlang.org/docs/interfaces.html

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/publishers/PrisonPersonEventPublisher.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    domainEventPublisher.publish(
+      DomainEvent(
+        eventType = eventType,
+        description = "A prison person record has been created",
+        detailUrl = detailUrl,
+        occurredAt = Instant.now().asStringWithUkZone(),
+        personReference = PersonReference(
+          identifiers = listOf(
+            PersonIdentifier("NOMS", noms),
```
markreesmoj: I know this is a refactor and it already uses this value, but is this right? NOMIS will be switched off in a year or so.
`prisonNumber` might be better? Anyone consuming it will know what we mean

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/publishers/PersonEventPublisher.kt
jldparker527: Yes wild! Think you can make default methods in Java these days as well.

I can see how using an interface for implementation could be useful as you can implement multiple interfaces and only extend one class. Given this class doesn't need multi inheritance, abstract base class seems more appropriate here though. But don't have a strong opinion either way

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/publishers/PrisonPersonEventPublisher.kt
lbevan: This was to be consistent with what we receive from events, though I don't believe there are any specifics on the ticket. If we think this is better then sure we can do that.

markreesmoj (dismissed):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/cprdomainevents/publishers/PrisonPersonEventPublisher.kt
lbevan: Update made to use `prisonNumber`

markreesmoj (approved):
👍

jldparker527 (approved):

PR #2155, opened 2026-06-24
CPR 1200 Simple cron job for Delius reconciliation

jldparker527:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


jldparker527:
Question: Will this be ok to run in preprod and prod. Do we wanna feature flag the config stuff?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/CorePersonRecordAndDeliusClient.kt
```diff
@@ -40,4 +43,29 @@ class CorePersonRecordAndDeliusClient(private val corePersonRecordAndDeliusWebCl
       .bodyToMono<ProbationAddress>()
       .block()!!,
   )
+
+  fun getProbationCases(params: CorePersonRecordAndDeliusClientPageParams): ProbationCases? = corePersonRecordAndDeliusWebClient
```
jldparker527, line 46: Don't think we will get a timeout this time here. Testing will confirm

helm_deploy/hmpps-person-record/values.yaml
```diff
@@ -150,5 +150,6 @@ generic-prometheus-alerts:
 
 cron:
   recordCountReportJob: "0 0-23 * * *"
+  deliusReconciliationReportJob: "0 7-17/2 * * 1-5"
```
markreesmoj: why not just run on the same schedule as the record count report?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/CorePersonRecordAndDeliusClient.kt
markreesmoj, line 46: :-) it's back, this has been in  and out of the client 4 or 5 times now

helm_deploy/hmpps-person-record/values.yaml
jldparker527: Fair point.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jobs/DeliusReconciliationReport.kt
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  suspend fun runDeliusReconciliation() {
+    val pageSize = 1
+    val totalPages = corePersonRecordAndDeliusClient.getProbationCases(
+      CorePersonRecordAndDeliusClientPageParams(
+        0,
+        1,
+      ),
+    )?.page?.totalPages ?: 1
```
markreesmoj: not sure we want to default this to 1 given what we are about to do with it. Maybe !! it instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jobs/DeliusReconciliationReport.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    val totalDeliusPersons = totalPages * pageSize
+    val totalCprPersons = personRepository.countBySourceSystemAndMergedToIsNullAndPassiveStateFalse(SourceSystemType.DELIUS)
+
+    applicationEventPublisher.publishEvent(
+      RecordTelemetry(
+        telemetryEventType = TelemetryEventType.CPR_RECORD_DELIUS_RECONCILIATION_REPORT,
+        elementMap = mapOf(
+          EventKeys.DELIUS to totalDeliusPersons.toString(),
+          EventKeys.CPR to totalCprPersons.toString(),
```
markreesmoj, line 52: shall we add a key saying whether they are different? It will make the event query and alert a bit easier to write
We haven't done that before because we haven't had the information available.
Alternatively we could ditch the call to our database and just publish the total from delius, then compare that to the total on the recordcountreport in AppInisghts

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jobs/DeliusReconciliationReport.kt
```diff
@@ -0,0 +1,57 @@
...
+  suspend fun runDeliusReconciliation() {
+    val pageSize = 1
+    val totalPages = corePersonRecordAndDeliusClient.getProbationCases(
+      CorePersonRecordAndDeliusClientPageParams(
+        0,
+        1,
+      ),
+    )?.page?.totalPages ?: 1
+
+    val totalDeliusPersons = totalPages * pageSize
```
markreesmoj, line 44: pagesize is always going to be 1 I think

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/CorePersonRecordAndDeliusClient.kt
```diff
@@ -40,4 +43,29 @@ class CorePersonRecordAndDeliusClient(private val corePersonRecordAndDeliusWebCl
...
+
+class CorePersonRecordAndDeliusClientPageParams(val page: Long, val size: Int) {
+  val sort: String = "id,asc"
+}
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class ProbationCases(
+  val page: PageMetadata,
+  @JsonProperty("content")
+  val cases: List<ProbationCase>,
```
markreesmoj: do we need to map the cases? We aren't doing anything with it I don't think

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/CorePersonRecordAndDeliusClient.kt
markreesmoj: I wouldn't worry about that either. It runs frequently enough for us to just let it fail every so often

markreesmoj:
> Question: Will this be ok to run in preprod and prod. Do we wanna feature flag the config stuff?

I think it will be fine - we will try it in dev first to make sure it works

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jobs/DeliusReconciliationReport.kt
jldparker527: Yes good catch

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/CorePersonRecordAndDeliusClient.kt
jldparker527: technically no. Will remove

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jobs/DeliusReconciliationReport.kt
jldparker527: We could do the first but feel like it's splitting hairs. Feel it would be nicer to see the number in telemetry

markreesmoj (approved):
👍

PR #2158, opened 2026-06-24
CPR-1228 Add types to the Probation merge and unmerge events

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/ProbationEvents.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/ProbationEvents.kt
@@ -1,6 +1,8 @@
 package uk.gov.justice.digital.hmpps.personrecord.client.model.sqs.messages.domainevent
 
 import com.fasterxml.jackson.annotation.JsonProperty
+import uk.gov.justice.digital.hmpps.personrecord.service.type.OFFENDER_MERGED
+import uk.gov.justice.digital.hmpps.personrecord.service.type.OFFENDER_UNMERGED
 
 data class ProbationOffenderDeleted(
   override val eventType: String,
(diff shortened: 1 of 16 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/ProbationEvents.kt
```diff
@@ -0,0 +1,36 @@
...
+
+data class ProbationOffenderMergedInfo(
+  @JsonProperty("sourceCRN")
+  val sourceCrn: String,
+
+  @JsonProperty("targetCRN")
+  val targetCrn: String,
+)
+
+data class ProbationOffenderUnMerged(
```
markreesmoj: We have considered `unmerged` to be a single word when casing variable names elsewhere in the codebase, I think we should probably do the same here.

markreesmoj: e.g. line 35 in this file

lbevan: Updated naming.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/ProbationEvents.kt
```diff
@@ -0,0 +1,36 @@
...
+
+data class ProbationOffenderDeleted(
+  override val eventType: String,
+  val personReference: PersonReference,
+) : HmppsDomainEvent {
+  val crn: String get() = personReference.identifiers?.first { it.type == "CRN" }?.value!!
+}
+
+data class ProbationOffenderMerged(
+  override val eventType: String,
```
markreesmoj:
```suggestion
  override val eventType: String = OFFENDER_MERGED,
```

Would this work? Would save supplying it in tests if it does work

lbevan: Nice simplification.

markreesmoj (approved):
👍

PR #2159, opened 2026-06-24
SDIT-3925: ✨ Map all old and new religion codes

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/types/ReligionCode.kt
```diff
@@ -1,38 +1,82 @@
...
-  SATN("Satanist"),
-  SCIE("Scientologist"),
-  SDAY("Seventh Day Adventist"),
-  SHIA("Muslim - Shia"),
-  SHNTAO("Shintoist and Taoist"),
-  SIKH("Sikh"),
-  SPIR("Spiritualist"),
-  TPNTS("They prefer not to say"),
-  ZORO("Zoroastrian"),
+enum class ReligionCode(val description: String, val current: Boolean) {
```
markreesmoj, line 3: Should we rename this PrisonReligionCode? Nothing else is using it - is it likely to?

petergphillips: In terms of the active ones, they are shared with Delius I believe.

petergphillips: and also ideally it would then be used by the [PersonEntity](https://github.com/ministryofjustice/hmpps-person-record/blob/b10102f236cb2efbc7f0abde990ce8df55e0fc6a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt#L110) for the religion as that currently uses the description from Prisoner Search but should really be using the enum.

markreesmoj: agreed

markreesmoj (approved):
👍

PR #2163, opened 2026-06-25
SDIT-3925: ✨ Switch prison religion to use new enum

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/prison/PrisonReligionEntity.kt
```diff
@@ -36,7 +37,7 @@ class PrisonReligionEntity(
   val prisonNumber: String,
 
   @Column(name = "religion_code", nullable = false)
-  var code: String,
+  var code: ReligionCode,
```
petergphillips, line 40: database column now uses the enum

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/types/ReligionCode.kt
```diff
@@ -78,9 +78,3 @@ enum class ReligionCode(val description: String, val current: Boolean) {
   WELS("Welsh Independent", false),
   ZORO("Zoroastrian", true),
 }
-
-fun String?.toReligionCodeDescription(): String? = try {
```
petergphillips, line 82: don't need this now as we retrieve using the enum instead

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/types/ReligionCodeTest.kt
```diff
@@ -1,24 +1,97 @@
...
-    val religionCodeEnum = ReligionCode.entries.random()
-    assertThat(religionCodeEnum.name.toReligionCodeDescription()).isEqualTo(religionCodeEnum.description)
-  }
-
-  @Test
-  fun `returns null when code is not recognised`() {
-    assertDoesNotThrow { assertThat("FAKE_OR_OLD_CODE".toReligionCodeDescription()).isEqualTo(null) }
+  @ParameterizedTest
+  @FieldSource("prisonReligionCodes")
+  fun `has a mapping for all of the NOMIS religion codes`(code: String) {
```
petergphillips, line 10: added in a test to ensure that all the (to be) migrated NOMIS religion codes can always be mapped okay

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/prison/PrisonReligion.kt
petergphillips, line 8: note that still using `String` at present for the api model. Will amend next to the `ReligionCode` enum

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/prison/PrisonReligionEntity.kt
markreesmoj: do we need 
 `@Enumerated(STRING) too?`

markreesmoj (approved):
👍

PR #2166, opened 2026-06-26
CPR-1232 prison update

JunaedAhmedMOJ:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
```diff
@@ -207,28 +207,7 @@ class PersonEntity(
...
-    this.disability = person.disability
-    this.immigrationStatus = person.immigrationStatus
-    this.birthplace = person.birthplace
-    this.birthCountryCode = person.birthCountryCode
-    this.nationalityNotes = person.nationalityNotes
-    this.updateChildEntities(person, childrenToIgnore)
-  }
-
-  private fun updateChildEntities(person: Person, childrenToIgnore: Set<KClass<*>>) {
+  fun updateChildEntities(person: Person, childrenToIgnore: Set<KClass<*>> = emptySet()) {
```
lbevan, line 210: Are we still considering removing `childrenToIgnore` from the call stack? Imagine a decent chunk of a change, maybe a separate PR if we do.

markreesmoj: yes, we are trying it now, will be a separate PR, you are right about it being a big change

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt
```diff
@@ -281,10 +260,6 @@ class PersonEntity(
 
   companion object {
 
-    fun new(person: Person, childrenToIgnore: Set<KClass<*>> = emptySet()): PersonEntity {
-      val personEntity = PersonEntity(sourceSystem = person.sourceSystem, matchId = UUID.randomUUID())
-      personEntity.update(person, childrenToIgnore)
-      return personEntity
-    }
+    fun new(person: Person): PersonEntity = PersonEntity(sourceSystem = person.sourceSystem, matchId = UUID.randomUUID())
```
lbevan: Only need `sourceSystem` from the person now. Maybe just pass that in so its clear its not using the whole person? Not sure of the impact of such a change.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/person/PersonEntityMappingTest.kt
```diff
@@ -59,7 +60,9 @@ class PersonEntityMappingTest {
   @Test
   fun `should update all entity fields from dto`() {
     // PersonEntity.new calls update which sets values from Person
-    val entity = PersonEntity.new(createPerson())
+    val person = createPerson()
+    val entity = PersonEntity.new(person)
+    entity.updatePersonEntity(person, emptySet())
```
markreesmoj: the `updatePersonEntity` does not return the personEntity so it can't be a one liner unless we change that, and nothing else requires it at the moment

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/person/PersonCreateAndUpdate.kt
```diff
@@ -0,0 +1,63 @@
...
+import kotlin.reflect.KClass
+
+fun PersonEntity.updatePersonEntity(
+  person: Person,
+  childrenToIgnore: Set<KClass<*>>,
+) {
+  when (this.sourceSystem) {
+    SourceSystemType.NOMIS -> fieldsToUpdatePrison(person)
+    else -> fieldsToUpdate(person, childrenToIgnore)
+  }
```
lbevan, line 16: Could return `this` here to make chaining easier. Fairly minor, but can be nice to reduce calls to 1 line in some places.

lbevan: Just seen your comment above. Fair 👍

lbevan (dismissed):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/person/PersonCreateAndUpdate.kt
markreesmoj: no your idea is better

lbevan (approved):
👍

PR #2168, opened 2026-06-26
CPR-1232 remove unused syscon sync controller

markreesmoj:
Nothing uses this nor is likely to any time soon
If we want it back, we revert this PR

Ref: https://github.com/ministryofjustice/hmpps-person-record/pull/2166
If we keep it, we will have to jump through a few hoops to keep it working
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


lbevan:
If it makes the other work simpler then its a fair thing to remove. Checked Azure and no requests to it 👍 . As you say, can revert if needed, though I guess Syscon will be doing the implementation, who know what changes will be needed by then. Fresh slate may help.

lbevan (dismissed):
👍

markreesmoj:
Thank goodness, I was just having a go at writing the update function for this and 🤼 💣 ⛑️ 🚧 👎

lbevan:
The diff is showing `SysconSyncControllerIntTest.kt` is no longer deleted and instead `SysconReligionControllerIntTest.kt` is. This seems wrong to me, but could be looking at it wrong.

PR #2170, opened 2026-06-26
CPR-1232 remove unused syscon sync controller

markreesmoj:
Nothing uses this nor is likely to any time soon
If we want it back, we revert this PR

Ref: https://github.com/ministryofjustice/hmpps-person-record/pull/2166
If we keep it, we will have to jump through a few hoops to keep it working

# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


lbevan (approved):
👍

PR #2175, opened 2026-06-30
CPR-1228 Add types to the prison domain events

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/PrisonEvents.kt
```diff
@@ -0,0 +1,33 @@
...
+
+data class PrisonPrisonerMerged(
+  override val eventType: String = PRISONER_MERGED,
+  val personReference: PersonReference,
+  val additionalInformation: PrisonPrisonerMergedInfo,
+) : HmppsDomainEvent {
+  val prisonNumber: String get() = personReference.identifiers?.first { it.type == "NOMS" }?.value!!
+}
+
+data class PrisonPrisonerMergedInfo(
```
markreesmoj: `PrisonPersonUpdated` might be better?
I can see we have already gone with `ProbationOffenderUpdated` etc - again I wonder if we should use `Person` instead of these terms?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/prison/PrisonEventListener.kt
```diff
@@ -16,9 +19,21 @@ class PrisonEventListener(
 ) {
 
   @SqsListener(Queues.PRISON_EVENT_QUEUE_ID, factory = "hmppsQueueContainerFactoryProxy")
-  fun onDomainEvent(rawMessage: String) = domainEventProcessor.process(rawMessage) {
-    prisonerSearchClient.getPrisoner(it.getPrisonNumber())?.let { person ->
+  fun onDomainEvent(rawMessage: String) = domainEventProcessor.processHmppsDomainEvent<HmppsDomainEvent>(rawMessage) { event ->
+    when (event) {
+      is PrisonPrisonerCreated -> processPrisonEvent(event.prisonNumber)
+      is PrisonPrisonerUpdated -> processPrisonEvent(event.prisonNumber)
```
markreesmoj: We have done this slightly differently in `ProbationEventListener` - was just looking at it this morning - where we have methods called (something like)`processCreateEvent` and `processUpdateEvent` which then simply call (something like) `processProbationEvent` 

I can see arguments for either but I think we should pick one way of doing it

markreesmoj (commented):
💬

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/prison/PrisonEventListener.kt
lbevan: I think this is simpler for now. I will update the probation one in the next PR, also have some cleanup/refactors in the same areas, at least this PR will be prison only.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/model/sqs/messages/domainevent/PrisonEvents.kt
lbevan: I'm happy either way, `Person` sounds fine.
I will update the prison ones here and do the probation ones separately.

For the the address ones like `ProbationPersonAddressUpdated` should we drop the `Person` from this and just have `ProbationAddressUpdated`?

markreesmoj (approved):
:+1

PR #2178, opened 2026-07-01
CPR-1232 tests

markreesmoj:
Start saving probation address details after creating the person 
This is to make reviewing the next PR easier, which will mean that addresses are no longer saved when creating a person

# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/IntegrationTestBase.kt
```diff
@@ -224,6 +225,13 @@ class IntegrationTestBase {
     dateOfBirth = randomDate(),
   )
 
+  internal fun addAddressToProbationRecord(initialAddress: uk.gov.justice.digital.hmpps.personrecord.model.person.Address): PersonEntity.() -> Unit = {
```
lbevan: The code suggests this is generic, i.e. can be used for any address, yet it is named specifically for Probation. I know we know that only probation bypasses address updates, but should we consider naming this generically?

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/config/IntegrationTestBase.kt
```diff
@@ -224,6 +225,13 @@ class IntegrationTestBase {
     dateOfBirth = randomDate(),
   )
 
+  internal fun addAddressToRecord(address: uk.gov.justice.digital.hmpps.personrecord.model.person.Address): PersonEntity.() -> Unit = {
```
jldparker527: Alias?

lbevan (approved):
👍

PR #2179, opened 2026-07-01
CPR-1228 Fix some probation events being discarded

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


lbevan:
Some events that are present in the CPE config but not in our code are being discarded in the ProbationEventListener event block since the typing changes. The simplest solution to this is to treat these event types as update events as all we do is update the whole person. It saves us having to create specific classes for each when we don't differentiate them. The signatures should match.

Said events:
- probation-case.engagement.recovered
- probation-case.alias.created
- probation-case.alias.updated
- probation-case.alias.deleted

I've used the updated event type naming convention (using `Person`). Others will get updated in the other PR.

We could consider throwing an exception instead of logging, so this type of issue is more apparent in future.

markreesmoj (approved):
👍

PR #2182, opened 2026-07-03
CPR-1232 no need to recluster here as personService will do it

markreesmoj:
This is the remand case creation code
Sequence of events:
**Probation Integration** consume a court message,
identify it is a remand case
check with CPR if we already have it
If not, create it
send us the details of the newly created probation record with the masterDefendantID

**CPR**
consumes this message, sets the masterDefendantID on the probation person and processes it.
There is a very high chance that we already have this probation record as it has been created in Delius and we will have consumed the person created event. 

MasterDefendantId is a matching field (currently although there are plans to remove this)
Consequently, since this has changed, it will be reclustered [here](https://github.com/ministryofjustice/hmpps-person-record/blob/62f5c4c9ec2449e5fea24effb31fd50332b96881/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/service/person/PersonService.kt#L54)

# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


lbevan (approved):
👍

jldparker527 (approved):

PR #2184, opened 2026-07-03
CPR-1201 Fix flaky assert when checking published CPR events

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


lbevan:
Would making the test queue fifo work/make sense in all scenarios?

lbevan:
We went this a more thorough solution to the actual problem of event potentially being published in the wrong order.
We decided to split the address handling into separate transactions, ensuring that the publishing happens in the order we expect. There may still be latency to aws, but at least our service respects the change of address order and should present them to PI correctly.

See #2186 for more info.

PR #2185, opened 2026-07-03
CPR-1234-change-zone-date

JunaedAhmedMOJ:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


jldparker527:
Not sure if we want to revert using ZonedDateTime everywhere. If we don't need ZondedDateTime, maybe we should change all field types for our core models back...

- `entity.AddressEntity`
- `model.person.Address`
- And also a migration script that that undos V197__ and V198__

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
```diff
@@ -13,12 +13,12 @@ data class CanonicalAddress(
   val noFixedAbode: Boolean? = null,
   @Schema(description = "Person address start date", example = "2020-02-26")
   val startDate: String? = null,
-  @Schema(description = "Person address start date time", example = "2026-02-26T11:08:46.347Z")
-  val startDateTime: ZonedDateTime? = null,
+  @Schema(description = "Person address start date time", example = "2026-02-26T11:08:46.347Z", type = "string", format = "date-time")
```
petergphillips: don't think you need the `type` or the `format`, should happen automatically

src/main/resources/db/migration/V201__modify_address_date_time.sql
```diff
@@ -0,0 +1,11 @@
...
+-------------------------------------------------------
+
+ALTER TABLE IF EXISTS personrecordservice.address
+    ALTER COLUMN start_date TYPE TIMESTAMP;
+
+ALTER TABLE IF EXISTS personrecordservice.address
+    ALTER COLUMN end_date TYPE TIMESTAMP;
+
+-----------------------------------------------------
+COMMIT;
```
lbevan, line 11: Should be fine, but lets check the conversion in dev to see that the midnight ones stay on the correct dates.

lbevan (approved):
👍

PR #2186, opened 2026-07-03
CPR-1201 Split out SAS arrived address handling into separate transactions

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


markreesmoj:
This should mean that PI get the main to previous change before the proposed to main change which is better for them as there is a maximum of 1 main address allowed in Delius

markreesmoj (approved):
👍

PR #2196, opened 2026-07-09
CPR-1242 release sas work to prod

jldparker527:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


jldparker527:
Do not release unless absolutely necessary!!!

markreesmoj (dismissed):
I do not recommend enabling this in production. It writes data to Delius and I am not confident that there has been sufficient testing in lower environments to prove that this will work as desired.
I am only approving as I am off on 10 July and it is likely that nobody else will be around to approve it tomorrow if we are instructed to put this live despite what I have said above

markreesmoj (approved):
👍

PR #2198, opened 2026-07-09
SDIT-3925: ✨ Workaround zoneddatetime issue by removing addresses from response

petergphillips:


markreesmoj:
Addresses do not use ZonedDateTime any more so no need for this workaround

PR #2202, opened 2026-07-13
CPR-1230-fix-swagger

JunaedAhmedMOJ:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


lbevan:
Is it possible to use the actual enum for the variable instead of a string? I'd imagine swagger generates the schema as a string and the default serialization uses `enum.name` anyways, which is the same output as we are currently doing.
If we can do that then we wouldn't have to specify our enum values twice.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalAddress.kt
```diff
@@ -37,7 +37,11 @@ data class CanonicalAddress(
   val county: String? = null,
   @Schema(description = "Person address country", example = "United Kingdom of Great Britain and Northern Ireland (the)")
   val country: String? = null,
-  @Schema(description = "Person address country code", example = "GBR")
+  @Schema(
```
markreesmoj, line 41: this is going to be impossible to maintain. Is there any way it can list the values of the enumeration that backs this field?

markreesmoj: Doesn't look like this is built in. Maybe something like this? https://stackoverflow.com/questions/38843911/passing-values-from-database-in-allowablevalues

markreesmoj (approved):
👍

PR #2193, opened 2026-07-09
Revert the revert for SDIT-3925: ✨ Switch prison religion to use new enum

petergphillips:
- **Reapply "SDIT-3925: ✨ Switch prison religion to use new enum (#2163)" (#2169)**
- **SDIT-3925: 🐛 Change code to be enum and test**

PR #2203, opened 2026-07-13
CPR-1233 writes and reads religion codes as enums for prison systems(nomis, syscon, dps)

jldparker527:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


jldparker527:
This PR should only really effect prison systems and the `prison_religion` table which is currently empty in PROD.

Next PR will be to update probation side

Final PR will be to update the PersonEntity.religion to be of type `ReligionCode`

lbevan:
Should get serialized as a string so don't think this should break any consumers (syscon only atm?).

markreesmoj:
Looks like this PR https://github.com/ministryofjustice/hmpps-person-record/pull/2193 does the same job. We approved and merged this one a couple of weeks ago but it was just missing the `@Enumerated` annotation on the entity, this is an attempt to reapply it

jldparker527:
> Should get serialized as a string so don't think this should break any consumers (syscon only atm?).

jldparker527:
> Should get serialized as a string so don't think this should break any consumers (syscon only atm?).

This change should effect calls that read/write into the `prison_relgion` table but no body is using them yet. Also as for the serialization, we will be storing the code as a ENUM so if it can't be saved anyway it's ok to not be serialized (fail fast)

But again, no body is using these endpoints yet as the `prison_religion` table is empty in PROD. Marius also informed me that everybody should be using these dedicated codes now

jldparker527:
> Looks like this PR #2193 does the same job. We approved and merged this one a couple of weeks ago but it was just missing the `@Enumerated` annotation on the entity, this is an attempt to reapply it

This PR only effect the reading of the values from the DB as an enum. Currently the table is empty. Also this ticket is for the saving of codes as enums aswell

jldparker527:
> > Looks like this PR #2193 does the same job. We approved and merged this one a couple of weeks ago but it was just missing the `@Enumerated` annotation on the entity, this is an attempt to reapply it
> 
> This PR only effect the reading of the values from the DB as an enum. Currently the table is empty. Also this ticket is for the saving of codes as enums aswell

Plan is above https://github.com/ministryofjustice/hmpps-person-record/pull/2203#issuecomment-4967051552

markreesmoj (approved):
👍

PR #2208, opened 2026-07-15
FIX: make less use of createPersonKey method

markreesmoj:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/jobs/ServiceNowMergeRequestE2ETest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/jobs/ServiceNowMergeRequestE2ETest.kt
@@ -238,12 +238,10 @@ class ServiceNowMergeRequestE2ETest : E2ETestBase() {
 
   @Test
   fun `should ignore merged records`() {
-    val person1 = createPerson(createRandomProbationPersonDetails())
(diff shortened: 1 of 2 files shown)


jldparker527:
Why this change? We use `createPersonKey` in many places

markreesmoj:
> Why this change? We use `createPersonKey` in many places

not sure - maybe we should keep it? I wanted to simplify the way we set up test data. I am not sure this is the right way to go. Maybe getting rid of `createPersonWithNewKey` would be a better option?

jldparker527:
> createPersonWithNewKey

Both `createPersonWithNewKey ` and `createPersonKey` are used 100+ places. If changing this has some benefit we should change it everywhere?

markreesmoj:
> > createPersonWithNewKey
> 
> Both `createPersonWithNewKey ` and `createPersonKey` are used 100+ places. If changing this has some benefit we should change it everywhere?

it is a bit of a mess - we generally just copy and paste when setting up data rather than pick the one which is most suitable.

Part of the problem is that we use these methods to set up data that realistically cannot exist - clusters with records which do not match for example. Ideally we would find a way to set up data that is more consistent with how it is created. Using API calls and messages as the E2E tests (largely) do is one way of achieving this, although the message setup in particular can be cumbersome.

PR #2212, opened 2026-07-15
CPR-1246 Recreate addresses when processing a Probation person recove…

lbevan:
…red event

# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/CorePersonRecordAndDeliusClient.kt
```diff
@@ -16,19 +16,23 @@ import uk.gov.justice.digital.hmpps.personrecord.service.queue.discardNotFoundEx
...
-    val probationCase = getProbationCase(crn)
+    val probationCase = fetchProbationCase(crn)
       .block()!!
     return Person.from(probationCase)
   }
 
-  private fun getProbationCase(crn: String): Mono<ProbationCase> = corePersonRecordAndDeliusWebClient
+  fun getProbationCase(crn: String): ProbationCase = fetchProbationCase(crn)
+    .discardNotFoundException()
+    .block()!!
```
lbevan, line 33: Seemed better to do this than map addresses in the core `Person` model. Happy to take views on that though.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationPersonRecoveredEventListenerIntTest.kt
```diff
@@ -0,0 +1,88 @@
...
+      val personEntity = personRepository.findByCrn(crn)
+      assertThat(personEntity).isNotNull()
+      assertThat(personEntity?.addresses?.size).isEqualTo(3)
+    }
+  }
+
+  @Test
+  fun `should update person and addresses if they already exist in cpr`() {
+    stubPersonMatchUpsert()
+    stubPersonMatchScores()
```
lbevan, line 39: Not expecting this to happen as the person would have been deleted, but if we somehow miss the event then CPR will update the person + addresses with the latest and grab any missing addresses.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/probation/ProbationPersonRecoveredEventProcessor.kt
```diff
@@ -0,0 +1,35 @@
...
+
+  @Transactional
+  fun process(event: ProbationPersonRecovered) {
+    corePersonRecordAndDeliusClient.getProbationCase(event.crn).let { case ->
+      val personEntity = probationProcessor.processProbationEvent(Person.from(case))
+      case.addresses.forEach { address ->
+        addressService.processAddress(
+          address = Address.from(address)!!,
+          findPerson = { personEntity },
+          findAddress = { addressRepository.findByDeliusAddressId(address.deliusAddressId) },
```
markreesmoj, line 29: I wonder what happens with the addressIds when a record is recovered? Would be interesting to try this out and see if they remain the same or are reassigned.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationPersonRecoveredEventListenerIntTest.kt
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  @Test
+  fun `should not recreate person if person does not exist in delius`() {
+    val crn = randomCrn()
+
+    stub404Response(probationUrl(crn))
+    publishProbationPersonRecoveredEvent(crn)
+
+    checkEventLogDoesNotExist(crn, CPRLogEvents.CPR_RECORD_CREATED)
```
markreesmoj, line 80: There is a race condition here. All these assertions are true at the start of the test and remain true so they pass immediately they are executed, which is possibly before the message consumption is complete. 

The thing that might change their value is the message consumption, but as it has no effect it is hard to see what we can assert.

Are we trying to prove that no exception is thrown when we get a 404?

I have checked the code out, removed `  .discardNotFoundException()` from `getProbationCase` and this test still passes

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationPersonRecoveredEventListenerIntTest.kt
```diff
@@ -0,0 +1,88 @@
...
+    checkEventLogDoesNotExist(crn, CPRLogEvents.CPR_RECORD_UPDATED)
+
+    awaitAssert {
+      val personEntity = personRepository.findByCrn(crn)
+      assertThat(personEntity).isNull()
+    }
```
markreesmoj, line 86:
```suggestion
    }
     expectNoMessagesOnQueueOrDlq(probationEventsQueue)
```

markreesmoj: I think this is a way of proving that we handle 404s without an exception

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/probation/ProbationPersonRecoveredEventProcessor.kt
lbevan: Hmm interesting. I guess they'd be new ids? 🤔 
Really this is just for it to work where it exists in CPR, though we expect the Person to have been deleted in most if not all cases. 
We could add code to identify when the Person exists and delete the addresses before adding the new ones.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationPersonRecoveredEventListenerIntTest.kt
lbevan: Fair point. The goal is to make sure if it doesn't exist in delius then we don't fail processing. I've added your suggestion below which helps identify that there is no message on the dlq in the end, though I agree the other asserts are a bit superfluous.

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationPersonRecoveredEventListenerIntTest.kt
lbevan: Added the suggestion. Thanks 👍

markreesmoj (dismissed):
👍

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/probation/ProbationPersonRecoveredEventProcessor.kt
markreesmoj: we have confirmed that a recovered address has the same deliusaddressid

JunaedAhmedMOJ (approved):

PR #2216, opened 2026-07-16
CPR-1233 Script to change old religion codes in our DB from Delius

jldparker527:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/resources/db/migration/V202_update_old_religion_codes.sql
+++ b/src/main/resources/db/migration/V202_update_old_religion_codes.sql
@@ -1,0 +1,13 @@
+BEGIN;
+-------------------------------------------------------
+
+UPDATE personrecordservice.person
+SET religion = 'UNKN'
+WHERE religion = '-1'
+
+UPDATE personrecordservice.person
+SET religion = 'TPRNTS'
+WHERE religion = 'REL01'
+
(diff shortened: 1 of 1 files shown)


markreesmoj:
Looks fine, but if I understand correctly any Delius updates to these records after this script is run will write `REL01` or `-1` back into the religion field and we will be back where we started

jldparker527:
> Looks fine, but if I understand correctly any Delius updates to these records after this script is run will write `REL01` or `-1` back into the religion field and we will be back where we started

They are legacy codes we shouldn't get them

src/main/resources/db/migration/V202_update_old_religion_codes.sql
```diff
@@ -0,0 +1,13 @@
+BEGIN;
```
JunaedAhmedMOJ, line 1: script number will need to be changed to 201

jldparker527:
Outstanding work - need to set up mapping of the old codes

https://mojdt.slack.com/archives/C061D3KQL92/p1784116309596699

lbevan:
Executed the scripts manually. This PR is not needed.

PR #2228, opened 2026-07-21
CPR-1233 Map old Delius religion codes

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/model/types/ReligionCode.kt
```diff
@@ -77,4 +77,11 @@ enum class ReligionCode(val description: String, val current: Boolean) {
   UR("United Reformed", false),
   WELS("Welsh Independent", false),
   ZORO("Zoroastrian", true),
+  ;
+
+  companion object {
+    fun fromProbation(code: String?): String? = PROBATION_RELIGION_CODE_MAPPING[code]?.let {
+      return it.name
+    }
+  }
```
lbevan, line 86: Still need to return a String as the PersonEntity hasn't been updated to use the enum yet.

markreesmoj (approved):
👍

PR #2235, opened 2026-07-23
FIX: only create a PersonEntity in tests where necessary

markreesmoj:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


lbevan:
Feels like `createRandomProbationPersonDetails` and `createRandomPrisonPersonDetails` functions should drop the `details` part of the name? Refering to Person I think is more appropriate since we are creating an instance of the `Person` model.

Just a thought, don't need to do it in this PR.

lbevan (approved):
👍

PR #2240, opened 2026-07-27
SDIT-4058: ✨ Allow insert to update existing religion entry

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/prison/PrisonReligionAPIController.kt
```diff
@@ -81,14 +82,13 @@ class PrisonReligionAPIController(
...
   )
   @PutMapping("/{prisonNumber}/religion/{cprReligionId}")
+  @ResponseStatus(HttpStatus.NO_CONTENT)
   fun updatePrisonReligion(
     @PathVariable("prisonNumber") prisonNumber: String,
     @PathVariable("cprReligionId") cprReligionId: String,
     @RequestBody requestBody: PrisonReligionUpdateRequest,
-  ): ResponseEntity<PrisonReligionSaveResponse> {
-    val prisonReligionMapping = prisonReligionUpdateHandler.handleUpdate(cprReligionId, requestBody)
-    val responseBody = PrisonReligionSaveResponse(prisonNumber, prisonReligionMapping)
```
petergphillips, line 90: we were never doing anything with the response anyway so no point returning it

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/prison/PrisonReligionInsertHandler.kt
```diff
@@ -31,6 +32,13 @@ class PrisonReligionInsertHandler(
     prisonReligionHistory: PrisonReligionHistory,
     personEntity: PersonEntity,
   ): String {
+    prisonReligionRepository.findByPrisonNumberAndCurrentPrisonRecordType(prisonNumber)?.let { existingCurrent ->
```
petergphillips, line 35: end date previous record

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/prison/PrisonReligionUpdateHandler.kt
```diff
@@ -3,47 +3,31 @@ package uk.gov.justice.digital.hmpps.personrecord.api.handler.prison
...
 
   @Transactional
-  fun handleUpdate(cprReligionId: String, updateRequest: PrisonReligionUpdateRequest): PrisonReligionMapping {
-    val existingPrisonReligion = validateRequest(cprReligionId, updateRequest)
-    val updatedPrisonReligion = updatePrisonReligion(updateRequest, existingPrisonReligion)
-    return PrisonReligionMapping(
-      nomisReligionId = updateRequest.nomisReligionId,
-      cprReligionId = updatedPrisonReligion.updateId.toString(),
-    )
+  fun handleUpdate(cprReligionId: String, updateRequest: PrisonReligionUpdateRequest) {
```
petergphillips, line 15: update processing will now only only be to change the comments of the existing row rather than to end date it

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/prison/PrisonReligionPostAPIControllerIntTest.kt
```diff
@@ -48,28 +48,35 @@ class PrisonReligionPostAPIControllerIntTest : WebTestBase() {
...
-    fun `person has existing current prison religion - does not save prison religion - return internal server error`() {
+    fun `person has existing current prison religion - saves new prison religion - updates current religion`() {
       val prisonNumber = randomPrisonNumber()
       val existingReligionEntity = PrisonReligionEntity.from(prisonNumber, createPrisonReligionHistory())
       val personEntityWithCurrentReligion = createPerson(createRandomPrisonPersonDetails(prisonNumber), configure = { religion = existingReligionEntity.code.name })
       prisonReligionRepository.save(existingReligionEntity)
       personRepository.saveAndFlush(personEntityWithCurrentReligion)
 
-      val requestBody = createPrisonReligionHistory()
+      val requestBody = createPrisonReligionHistory().copy(startDate = existingReligionEntity.startDate.plusDays(1))
```
petergphillips, line 58: need to ensure that the new history entry is created with a start date after the original otherwise they end up being returned in the wrong order

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/prison/PrisonReligionPostAPIControllerIntTest.kt
```diff
@@ -48,28 +48,35 @@ class PrisonReligionPostAPIControllerIntTest : WebTestBase() {
...
       sendPostRequestAsserted<Unit>(
         url = prisonReligionPostEndpoint(prisonNumber),
         body = requestBody,
         roles = listOf(PERSON_RECORD_SYSCON_SYNC_WRITE),
-        expectedStatus = HttpStatus.INTERNAL_SERVER_ERROR,
+        expectedStatus = HttpStatus.CREATED,
       )
 
-      awaitAssert {
```
petergphillips, line 66: no idea why there was an `awaitAssert` here. The transaction will have been committed by the post request anyway

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/prison/PrisonReligionRepository.kt
```diff
@@ -9,5 +10,8 @@ import java.util.UUID
 interface PrisonReligionRepository : JpaRepository<PrisonReligionEntity, Long> {
   fun deleteAllByPrisonNumber(prisonNumber: String)
   fun findByPrisonNumberOrderByStartDateDescCreateDateTimeDesc(prisonNumber: String): List<PrisonReligionEntity>
+
+  @Query("SELECT r FROM PrisonReligionEntity r WHERE r.prisonNumber = :prisonNumber AND r.prisonRecordType = 'CURRENT'")
+  fun findByPrisonNumberAndCurrentPrisonRecordType(prisonNumber: String): PrisonReligionEntity?
   fun findByUpdateId(updateId: UUID): PrisonReligionEntity?
```
lbevan, line 16: I think we could use the following to avoid the custom query:
`fun findByPrisonNumberAndPrisonRecordType(prisonNumber: String, prisonRecordType: PrisonRecordType): PrisonReligionEntity?`
Though its fairly minor given we only use it in 1 place :)

lbevan (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/repository/prison/PrisonReligionRepository.kt
petergphillips: I did think of that, but we would only ever want to search for `CURRENT` anyway. Searching for `HISTORIC` we would want a list since there could be more than one. In the past I've written two repository methods, one that then calls the other with `CURRENT` for the type but that has always felt naff as a way to avoid the custom query

lbevan: Ahh yeah fair point 👍

PR #2244, opened 2026-07-29
CPR-1197 Retry on 404 when calling Delius for person details

JunaedAhmedMOJ:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/listeners/probation/ProbationEventListenerIntTest.kt
```diff
@@ -711,11 +712,32 @@ class ProbationEventListenerIntTest : ProbationEventListenerTestBase() {
...
+    stubPersonMatchScores()
+    stubGetRequest(
+      scenarioName = "retry-404",
+      currentScenarioState = "Started",
+      nextScenarioState = "request-succeeded",
+      url = probationUrl(crn),
+      body = "",
+      status = 404,
+    )
+    stubGetRequest(
```
jldparker527, line 735: Does this definitely force a 404 then a 200 and not just override the stub mappings?

JunaedAhmedMOJ: yes retry-404 scenario would do this as it forces a 404 then a 200

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/CorePersonRecordAndDeliusClient.kt
```diff
@@ -15,9 +18,13 @@ import uk.gov.justice.digital.hmpps.personrecord.service.queue.discardNotFoundEx
 @Component
 class CorePersonRecordAndDeliusClient(private val corePersonRecordAndDeliusWebClient: WebClient) {
 
+  @Retryable(
+    maxAttempts = 5,
+    backoff = Backoff(delay = 200, random = true, multiplier = 3.0),
+    retryFor = [WebClientResponseException.NotFound::class],
+  )
```
lbevan: Is it possible to use `CprRetryable` here instead and just pass the `retryFor` value? All the other inputs are defaulted in our version I think.

lbevan: This is currently only applied to the `getPerson` call but there other ones this may make sense to use too, like `getAddress`. Review the other public functions in this client to see if its useful to add for them.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/CorePersonRecordAndDeliusClient.kt
lbevan, line 31: These functions are now identical, except for the retry. What is the use case for each?

JunaedAhmedMOJ: getPersonErrorIfNotFound is used specifically on unmerged and we do not want to do retries for this and we do not want it to go to DLQ

lbevan: Pretty sure this is the other way around? i.e. if we get a 404 on unmerge we want to fail and go to the DLQ.
Since we have removed `.discardNotFoundException()` from the other function, it is functionally the same except for the retry, but I seems to me this is only a benefit for the unmerge scenario?

JunaedAhmedMOJ: yes sorry i meant when the getPersonErrorIfNotFound gets a 404 it will go to DLQ

JunaedAhmedMOJ: From the ticket description i can see that there was an issue with retries not happening for 'probation-case.address.created'

lbevan: I think we can probably add it to the other remaining functions and remove getPersonErrorIfNotFound , but we can do that separately

lbevan (approved):
👍

PR #2248, opened 2026-07-31
CPR-1233 added endpoint to map old religion codes to new ones

jldparker527:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/admin/ReligionCodeMappingControllerIntTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/admin/ReligionCodeMappingControllerIntTest.kt
@@ -1,0 +1,36 @@
+package uk.gov.justice.digital.hmpps.personrecord.api.controller.admin
+
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+import org.springframework.http.HttpStatus
+import uk.gov.justice.digital.hmpps.personrecord.config.WebTestBase
+import uk.gov.justice.digital.hmpps.personrecord.model.types.ReligionCode
+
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/admin/ReligionCodeMappingControllerIntTest.kt
```diff
@@ -0,0 +1,36 @@
...
+
+  @Test
+  fun `maps all old codes to new codes`() {
+    val cluster = createPersonKey()
+    ReligionCode.entries.forEach {
+      val person = createRandomProbationPersonDetails().copy(religion = it.description)
+      val personEntity = createPerson(person)
+      personEntity.personKey = cluster
+      cluster.personEntities.add(personEntity)
+      personRepository.save(personEntity)
```
jldparker527, line 19: cluster.addPerson doesn't work here because of `org.springframework.orm.ObjectOptimisticLockingFailureException: Row was already updated or deleted by another transaction for entity`

jldparker527:
Temporary endpoint to map the remaining 1,174 religion codes to the new ones

Note - the remaining are ALL merged records

lbevan:
> Temporary endpoint to map the remaining 1,174 religion codes to the new ones
> 
> Note - the remaining are ALL merged records

There are `1190` using the query in the PR, which includes all records not just NOMIS. Is this intended? I assume so as we'll need to update them anyhow, but was there some discussion on doing something different? I can't remember.

lbevan:
We could be a bit more careful with the religion lookup, falling back to `null` if the religion code is not found. Not sure if we expect any of these but would be a small change and would avoid an additional merge and deploy if we did find any.

lbevan (approved):
👍

PR #2257, opened 2026-08-03
CPR-1233 handles alt codes

jldparker527:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/admin/ReligionCodeMappingController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/admin/ReligionCodeMappingController.kt
@@ -18,13 +18,23 @@ class ReligionCodeMappingController(
 
   private val religionCodesByOldCodes = ReligionCode.entries.associateBy { it.description }
 
+  private val alternativeCodesByNewCodes = mapOf(
+    "Christian" to ReligionCode.CHRST,
+    "Church of England (Anglican)" to ReligionCode.COFE,
+    "Roman Catholic" to ReligionCode.RC,
+    "No Religion" to ReligionCode.NIL,
+  )
+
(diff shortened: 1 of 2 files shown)


jldparker527:
See thread for context https://mojdt.slack.com/archives/C054W4V32R2/p1785764287735649

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/admin/ReligionCodeMappingController.kt
```diff
@@ -18,13 +18,23 @@ class ReligionCodeMappingController(
...
   suspend fun mapOldReligionCodes() {
     CoroutineScope(Dispatchers.Default).launch {
       val personsWithOldReligionCodes = personRepository.findAllPersonsWithOldReligionCodes()
-      personsWithOldReligionCodes.forEach {
-        it.religion = religionCodesByOldCodes.getOrDefault(it.religion, null)?.name
+      personsWithOldReligionCodes.forEach { personEntity ->
+        val religionCode = religionCodesByOldCodes[personEntity.religion] ?: alternativeCodesByNewCodes[personEntity.religion]
+        if (religionCode != null) {
+          personEntity.religion = religionCode.name
+        }
```
jldparker527, line 37: Simply just skip and not set to null. Should not be any left but in the very unlikely event, I would like to re-run the query and see why

lbevan (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/admin/ReligionCodeMappingController.kt
lbevan: Can resolve manually if a low amount left 👍

PR #2272, opened 2026-08-07
SDIT-4058 Merge religion history

richardpopple:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/prison/PrisonReligionMergeHandler.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/prison/PrisonReligionMergeHandler.kt
@@ -1,25 +1,27 @@
 package uk.gov.justice.digital.hmpps.personrecord.api.handler.prison
 
+import org.springframework.context.ApplicationEventPublisher
 import org.springframework.stereotype.Component
 import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.PersonEntity
 import uk.gov.justice.digital.hmpps.personrecord.jpa.entity.prison.PrisonReligionEntity
+import uk.gov.justice.digital.hmpps.personrecord.jpa.repository.PersonRepository
 import uk.gov.justice.digital.hmpps.personrecord.jpa.repository.prison.PrisonReligionRepository
-import uk.gov.justice.digital.hmpps.personrecord.model.person.Person
 import uk.gov.justice.digital.hmpps.personrecord.model.types.PrisonRecordType.CURRENT
 import uk.gov.justice.digital.hmpps.personrecord.model.types.PrisonRecordType.HISTORIC
-import uk.gov.justice.digital.hmpps.personrecord.service.person.PersonService
+import uk.gov.justice.digital.hmpps.personrecord.service.cprdomainevents.events.religion.PrisonPersonReligionsMerged
 import java.time.LocalDate
 
 @Component
 class PrisonReligionMergeHandler(
   private val prisonReligionRepository: PrisonReligionRepository,
-  private val personService: PersonService,
+  private val personRepository: PersonRepository,
+  private val publisher: ApplicationEventPublisher,
 ) {
 
   @Transactional
-  fun handleMerge(from: PersonEntity?, to: PersonEntity) {
-    if (from?.prisonNumber == null || to.prisonNumber == null) {
+  fun handleMerge(from: PersonEntity?, to: PersonEntity?) {
+    if (from?.prisonNumber == null || to?.prisonNumber == null) {
       return
     }
     val fromHistory: List<PrisonReligionEntity> = prisonReligionRepository.findByPrisonNumberOrderByStartDateDescCreateDateTimeDesc(from.prisonNumber!!)
@@ -34,10 +36,11 @@ class PrisonReligionMergeHandler(
       toBeHistoric.endDate = LocalDate.now()
       prisonReligionRepository.saveAndFlush(toBeHistoric)
       to.religion = current.code
-      personService.processPerson(Person.from(to)) { to }
+      personRepository.saveAndFlush(to)
     }
 
     // Put all of the religions on the to person
     fromHistory.forEach { it.prisonNumber = to.prisonNumber!! }
+    publisher.publishEvent(PrisonPersonReligionsMerged(from, to))
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/prison/PrisonReligionMergeHandler.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/prison/PrisonReligionMergeHandler.kt
@@ -2,14 +2,10 @@ package uk.gov.justice.digital.hmpps.personrecord.api.handler.prison
 
 import org.springframework.stereotype.Component
 import org.springframework.transaction.annotation.Transactional
-import uk.gov.justice.digital.hmpps.personrecord.api.controller.exceptions.ResourceNotFoundException
-import uk.gov.justice.digital.hmpps.personrecord.api.model.prison.PrisonReligionMapping
-import uk.gov.justice.digital.hmpps.personrecord.api.model.sysconsync.historic.PrisonReligionHistory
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/prison/PrisonReligionMergeHandler.kt
```diff
@@ -0,0 +1,42 @@
...
+    val fromHistory: List<PrisonReligionEntity> = prisonReligionRepository.findByPrisonNumberOrderByStartDateDescCreateDateTimeDesc(from.prisonNumber!!)
+    val toHistory: List<PrisonReligionEntity> = prisonReligionRepository.findByPrisonNumberOrderByStartDateDescCreateDateTimeDesc(to.prisonNumber!!)
+
+    // Choose the current religion with the latest start date and set the religion on the to person
+    val currentReligions = (fromHistory + toHistory).filter { it.prisonRecordType == CURRENT }
+    if (currentReligions.size > 1) {
+      val toBeHistoric = currentReligions.minBy { it.startDate }
+      val current = currentReligions.maxBy { it.startDate }
+      toBeHistoric.prisonRecordType = HISTORIC
+      toBeHistoric.endDate = current.startDate
```
petergphillips: not sure about this, Prison API uses `now()` when it sets the end date

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/prison/PrisonReligionMergeHandler.kt
```diff
@@ -0,0 +1,42 @@
...
+      return
+    }
+    val fromHistory: List<PrisonReligionEntity> = prisonReligionRepository.findByPrisonNumberOrderByStartDateDescCreateDateTimeDesc(from.prisonNumber!!)
+    val toHistory: List<PrisonReligionEntity> = prisonReligionRepository.findByPrisonNumberOrderByStartDateDescCreateDateTimeDesc(to.prisonNumber!!)
+
+    // Choose the current religion with the latest start date and set the religion on the to person
+    val currentReligions = (fromHistory + toHistory).filter { it.prisonRecordType == CURRENT }
+    if (currentReligions.size > 1) {
+      val toBeHistoric = currentReligions.minBy { it.startDate }
+      val current = currentReligions.maxBy { it.startDate }
```
petergphillips: what happens if the `startDate` is the same on both records, could `toBeHistoric` and `current` actually then be the same record?

petergphillips: actually the first entry in `fromHistory` or `toHistory` is the only one that could be current (not sure it is guaranteed to be though)

lbevan:
I guess this is related to the thread in slack https://mojdt.slack.com/archives/C084T9701DX/p1786098360962909
Would be good to understand why this logic needs to live in CPR - is it not possible to do this via the existing API calls?

Is `PrisonReligionMergeHandler` intended to be called from `PrisonMergeEventProcessor`? 

EDIT: we had a chat, more to think about next week but we are on the right track!

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/prison/PrisonReligionMergeHandler.kt
```diff
@@ -0,0 +1,43 @@
...
+      val current = currentReligions[1]
+      toBeHistoric.prisonRecordType = HISTORIC
+      toBeHistoric.endDate = LocalDate.now()
+      prisonReligionRepository.saveAndFlush(toBeHistoric)
+      to.religion = current.code
+      personService.processPerson(Person.from(to)) { to }
+    }
+
+    // Put all of the religions on the to person
+    fromHistory.forEach { it.prisonNumber = to.prisonNumber!! }
```
lbevan, line 43: Being explicit with a JPA save/saveall here may be clearer.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/prison/PrisonReligionMergeHandler.kt
richardpopple: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/prison/PrisonReligionMergeHandler.kt
```diff
@@ -0,0 +1,46 @@
...
+      toBeHistoric.prisonRecordType = HISTORIC
+      toBeHistoric.endDate = LocalDate.now()
+      prisonReligionRepository.saveAndFlush(toBeHistoric)
+      to.religion = current.code
+      personRepository.saveAndFlush(to)
+    }
+
+    // Put all of the religions on the to person
+    fromHistory.forEach { it.prisonNumber = to.prisonNumber!! }
+    publisher.publishEvent(PrisonPersonReligionsMerged(from, to))
```
richardpopple: Publish the merge religion event here. If however we wanted a slightly more generic a "nomis merge has happened and cpr needed to do some additional actions" event then that should probably be raised at the end of the PrisonMergeEventProcessor code

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/prison/PrisonReligionMergeHandler.kt
```diff
@@ -0,0 +1,46 @@
...
+    // Choose the current religion with the latest start date and set the religion on the to person
+    val currentReligions = (fromHistory + toHistory).filter { it.prisonRecordType == CURRENT }.sortedBy { it.startDate }
+    if (currentReligions.size == 2) {
+      val toBeHistoric = currentReligions[0]
+      val current = currentReligions[1]
+      toBeHistoric.prisonRecordType = HISTORIC
+      toBeHistoric.endDate = LocalDate.now()
+      prisonReligionRepository.saveAndFlush(toBeHistoric)
+      to.religion = current.code
+      personRepository.saveAndFlush(to)
```
richardpopple, line 39: As part of the merge of the religious history we need to update the religion on the person entity too.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/prison/PrisonMergeEventProcessor.kt
```diff
@@ -15,15 +16,18 @@ class PrisonMergeEventProcessor(
...
 ) {
 
   @Transactional
   fun processEvent(domainEvent: PrisonPersonMerged) {
     val prisonNumber = domainEvent.prisonNumber
     prisonerSearchClient.getPrisoner(prisonNumber)?.let {
       val from: PersonEntity? = personRepository.findByPrisonNumber(domainEvent.additionalInformation.sourcePrisonNumber)
-      val to: PersonEntity = personService.processPerson(it.doNotReclusterOnUpdate()) { personRepository.findByPrisonNumber(prisonNumber) }
-      mergeService.processMerge(from, to)
+      val to = personRepository.findByPrisonNumber(prisonNumber)
```
richardpopple, line 27: Before we do any processing of the person entity we merge the religion history. Note that the processing of the person entity ignores the religion. By merging the religion here we do so before raising any down stream process events, and while this actually doesn't make much difference as the listeners are all @transactionEvents in maybe makes it a bit more explicit

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/message/processors/prison/PrisonMergeEventProcessor.kt
```diff
@@ -15,15 +16,18 @@ class PrisonMergeEventProcessor(
   private val mergeService: MergeService,
   private val prisonerSearchClient: PrisonerSearchClient,
   private val personService: PersonService,
+  private val prisonReligionMergeHandler: PrisonReligionMergeHandler,
 ) {
 
   @Transactional
   fun processEvent(domainEvent: PrisonPersonMerged) {
```
richardpopple, line 23: Ultimately we'd like from nomis to listen for this event and call an endpoint

PR #2274, opened 2026-08-07
CPR-1216 Consolidates all known identifiers for single PersonEntity, not the cluster

jldparker527:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalIdentifiers.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalIdentifiers.kt
@@ -119,6 +119,6 @@ data class CanonicalIdentifiers(
       )
     }
 
-    private fun List<ReferenceEntity>.findByIdentifierType(identifierType: IdentifierType): List<String> = this.filter { it.identifierType == identifierType }.mapNotNull { it.identifierValue }
+    fun List<ReferenceEntity>.findByIdentifierType(identifierType: IdentifierType): List<String> = this.filter { it.identifierType == identifierType }.mapNotNull { it.identifierValue }
   }
 }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/search/PersonSearchResponse.kt
```diff
@@ -36,35 +41,39 @@ data class SearchData(
...
-        identifiers = personEntity.references.map { SearchIdentifier.from(it) },
+        identifiers = CanonicalIdentifiers.forSinglePerson(personEntity),
         sourceSystem = personEntity.sourceSystem,
         status = personEntity.personKey!!.status.toSearchStatus(),
       )
     }
   }
 }
 
+fun CanonicalIdentifiers.Companion.forSinglePerson(personEntity: PersonEntity): CanonicalIdentifiers {
```
lbevan: Seems like this is functionally the same as `fun from(personEntities: List<PersonEntity>)` in `CanonicalIdentifiers`? Unless im missing something. Can we reuse that but just pass 1 entity to it in a list?

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/canonical/CanonicalIdentifiers.kt
```diff
@@ -12,6 +13,7 @@ import uk.gov.justice.digital.hmpps.personrecord.model.types.IdentifierType.NATI
 import uk.gov.justice.digital.hmpps.personrecord.model.types.IdentifierType.OTHR
 import uk.gov.justice.digital.hmpps.personrecord.model.types.IdentifierType.PNC
 
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
```
jldparker527: If we keep the latest commit, identifiers will look like this in the response...

`
"identifiers": {
  "prisonNumber": "KG4448L",
  "pncs": [
    "2022/7786175T"
  ],
  "cros": [
    "335945/79X"
  ]
},
`

Idea behind it is, personEntity can only have one prisonNumber, crn, defendantId and cid so they shouldn't be a list in the response. As for the other Identifiers, a personEntity can have multiple of the same

lbevan (approved):
👍

PR #2277, opened 2026-08-10
CPR-1216 Change request body on CPR side and release to PREPROD

jldparker527:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/search/PersonSearchRequest.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/search/PersonSearchRequest.kt
@@ -2,22 +2,20 @@ package uk.gov.justice.digital.hmpps.personrecord.api.model.search
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties
 import io.swagger.v3.oas.annotations.media.Schema
-import jakarta.validation.constraints.NotBlank
 import java.time.LocalDate
 
 @JsonIgnoreProperties(ignoreUnknown = true)
+@Schema(name = "PersonSearchRequest")
 data class PersonSearchRequest(
   @Schema(required = true)
-  @NotBlank
   val firstName: String,
   @Schema(required = true)
-  @NotBlank
   val lastName: String,
   val middleName: String? = null,
   @Schema(required = true)
   val dateOfBirth: LocalDate,
-  val firstNameAliases: List<String> = emptyList(),
-  val lastNameAliases: List<String> = emptyList(),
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/search/PersonSearchRequest.kt
```diff
@@ -1,11 +1,20 @@
...
 data class PersonSearchRequest(
-  val fullName: String,
+  @Schema(required = true)
+  @NotBlank
+  val firstName: String,
+  @Schema(required = true)
+  @NotBlank
+  val lastName: String,
+  val middleName: String? = null,
+  @Schema(required = true)
```
lbevan, line 15: Are Tony and co OK with these changes?

The only issue I can see is if Vetting/EMDI only have partial data and cannot fill in some of these now mandatory fields, though we will have to see the feedback from them I guess.
Is `dateOfBirth` actually required here?

jldparker527: Simply just made this match what match accepts. Given Person Record will be the front door, I have made the request body easier to understand (firstName, lastName and DOB) are mandatory from match side.

Not ok'd with Tony...

jldparker527: Spoke with Amol. All good. We are only changing our request layout

jldparker527:
Screenshot 2026-08-10 at 15 52 17
Request will know look like this for users

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/model/search/PersonSearchRequest.kt
```diff
@@ -1,14 +1,21 @@
...
+  val firstName: String,
+  @Schema(required = true)
+  val lastName: String,
+  val middleName: String? = null,
+  @Schema(required = true)
   val dateOfBirth: LocalDate,
-  val firstNameAliases: List<String> = emptyList(),
-  val lastNameAliases: List<String> = emptyList(),
-  val dateOfBirthAliases: List<LocalDate> = emptyList(),
-  val postcodes: List<String> = emptyList(),
```
jldparker527, line 13: Swagger makes these fields *mandatory if I keep them as defaulting to empty list

lbevan (approved):
👍

lbevan (approved):
👍

PR #2282, opened 2026-08-12
CPR-1259 Add Common Platform candidate matches endpoint

lbevan:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/search/CandidateSearchHandler.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/search/CandidateSearchHandler.kt
@@ -1,0 +1,25 @@
+package uk.gov.justice.digital.hmpps.personrecord.api.handler.search
+
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.personrecord.api.model.canonical.CanonicalRecord
+import uk.gov.justice.digital.hmpps.personrecord.client.PersonMatchClient
+import uk.gov.justice.digital.hmpps.personrecord.jpa.repository.PersonRepository
+import uk.gov.justice.digital.hmpps.personrecord.model.types.SourceSystemType
+import java.util.UUID
+
+@Component
+class CandidateSearchHandler(
+  private val personRepository: PersonRepository,
+  private val personMatchClient: PersonMatchClient,
+) {
+
+  fun searchByDefendantId(defendantId: String): List<CanonicalRecord> = personRepository.findByDefendantId(defendantId)
+    ?.let { getCandidateMatches(it.matchId) } ?: emptyList()
+
+  private fun getCandidateMatches(matchId: UUID): List<CanonicalRecord> = getCandidateMatchScoresSortedByMatchWeightDescending(matchId)
+    .map { personRepository.findByMatchId(UUID.fromString(it.candidateMatchId))!! }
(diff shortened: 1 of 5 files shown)


lbevan:
Planning to add E2E tests once the person-match endpoint is available.

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/search/CandidateSearchHandler.kt
```diff
@@ -0,0 +1,25 @@
...
+  private val personMatchClient: PersonMatchClient,
+) {
+
+  fun searchByDefendantId(defendantId: String): List<CanonicalRecord> = personRepository.findByDefendantId(defendantId)
+    ?.let { getCandidateMatches(it.matchId) } ?: emptyList()
+
+  private fun getCandidateMatches(matchId: UUID): List<CanonicalRecord> = getCandidateMatchScoresSortedByMatchWeightDescending(matchId)
+    .map { personRepository.findByMatchId(UUID.fromString(it.candidateMatchId))!! }
+    .filter { it.sourceSystem == SourceSystemType.DELIUS }
+    .map { CanonicalRecord.from(it) }
```
markreesmoj, line 22: if we have a definite match, should we just return that?

lbevan: The flow, as I gather, is that they call this endpoint if there are no delius records on the same cluster as the common platform record. So given that, will there ever be a "definite" match from this new endpoint? All will be under the 20 match threshold and wont be on the cluster with the common platform record.
Unless my understanding of a definite match is wrong here. Lmk!

markreesmoj: The thing is that PiC do not check whether there is a match (yet). It is possible that PiC will write a batch job which will check whether there is a match on a daily (?) basis. However that is not there yet, and even when it is will probably not cover all matches. 

So I think it very likely that there will be definite matches in these results sometimes. If it were in use now, we would expect definite matches more often than not based on the manual linking stats which show about 60% of manual links are already in place

lbevan: Okay. I guess I'm a bit confused as to what the ticket means then, as it says they already do a call for the cluster

> Prepare a Case for Sentence make their usual call to CPR to see if we hold any Delius records on the cluster the court record sits on. If there are no Delius records on the cluster they call us back on a different endpoint to see if we can find any ‘possible’ Delius matches for their court record. 

So are we saying first check if there has already been a match and the record is on a cluster with a Delius record(s). If they are then return those as the list first.
If there aren't any Delius records on the cluster then do this new search and return a match weight sorted list of Delius records.

markreesmoj: we should discuss this with PaCFS developers and see what they want
If we know which CRN a defendantId matches, why should we present a list to a user to pick from?

lbevan:
Closing as this is on hold. PiC are first going to call us to get updated matches. That is the first step before using our data to inform matching.

PR #2287, opened 2026-08-13
CPR-1238 added previous_court_address table

jldparker527:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/resources/db/migration/V201__create_previous_court_address_table.sql
+++ b/src/main/resources/db/migration/V201__create_previous_court_address_table.sql
@@ -1,0 +1,18 @@
+BEGIN;
+-------------------------------------------------------
+
+CREATE TABLE IF NOT EXISTS personrecordservice.previous_court_addresss
+(
+    id                                      SERIAL PRIMARY KEY,
+    fk_person_id                            BIGINT NOT NULL,
+    building_name                           TEXT NULL,
+    building_number                         TEXT NULL,
+    thoroughfare_name                       TEXT NULL,
+    dependent_locality                      TEXT NULL,
+    post_town                               TEXT NULL,
+    postcode                                TEXT NULL,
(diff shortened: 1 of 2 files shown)


src/main/resources/db/migration/V201__create_previous_court_address_table.sql
```diff
@@ -0,0 +1,18 @@
...
+CREATE TABLE IF NOT EXISTS personrecordservice.previous_court_addresss
+(
+    id                                      SERIAL PRIMARY KEY,
+    fk_person_id                            BIGINT NOT NULL,
+    building_name                           TEXT NULL,
+    building_number                         TEXT NULL,
+    thoroughfare_name                       TEXT NULL,
+    dependent_locality                      TEXT NULL,
+    post_town                               TEXT NULL,
+    postcode                                TEXT NULL,
```
markreesmoj, line 13: we only need postcode for matching - have we a requirement for storing all of this other additional data?

jldparker527: These fields are listed on the ticket https://dsdmoj.atlassian.net/browse/CPR-1238

markreesmoj: yes, but do we actually need them, or are they simply copied and pasted from the address table (where we do need them)?

lbevan: If there are no plans to use other address fields for matching, then yeah we can consider just having postcodes here.

jldparker527: Spoke with Tony and he's going to get back to me. He wan't to double check if we have anything on the horizon that would result in us wanting these fields

jldparker527:
Going with a different approach

PR #2288, opened 2026-08-13
CPR-1107 refactoring

markreesmoj:
# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/CorePersonRecordAndDeliusClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/CorePersonRecordAndDeliusClient.kt
@@ -12,7 +12,6 @@ import uk.gov.justice.digital.hmpps.personrecord.client.model.offender.Probation
 import uk.gov.justice.digital.hmpps.personrecord.client.model.offender.ProbationCase
 import uk.gov.justice.digital.hmpps.personrecord.model.person.Address
 import uk.gov.justice.digital.hmpps.personrecord.model.person.Person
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/CorePersonRecordAndDeliusClient.kt
```diff
@@ -24,15 +23,7 @@ class CorePersonRecordAndDeliusClient(private val corePersonRecordAndDeliusWebCl
     return Person.from(probationCase)
   }
 
-  fun getPersonErrorIfNotFound(crn: String): Person {
```
markreesmoj, line 27: this is identical to the above `getPerson` method apart from the retry

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/client/CorePersonRecordAndDeliusClient.kt
```diff
@@ -24,15 +23,7 @@ class CorePersonRecordAndDeliusClient(private val corePersonRecordAndDeliusWebCl
...
-    val probationCase = fetchProbationCase(crn)
-      .block()!!
-    return Person.from(probationCase)
-  }
-
-  fun getProbationCase(crn: String): ProbationCase = fetchProbationCase(crn)
-    .discardNotFoundException()
-    .block()!!
```
markreesmoj, line 35: used when recovering a Delius record - no reason to discard a 404 here - it would be an error we should investigate otherwise we would end up out of sync with Delius

lbevan (approved):
👍

PR #2289, opened 2026-08-13
CPR-1197 Discard 404 exceptions after retries

lbevan:
This is now needed because in dev Delius tests delete records without sending out events.
In practice, we don't get many 404's in production, so this would very rarely happen. We have enhanced logging to help detect when this happens.

# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged


markreesmoj (dismissed):
👍

markreesmoj (dismissed):
👍

markreesmoj (approved):
👍

PR #2291, opened 2026-08-14
CPR-1260-remand-link-process

JunaedAhmedMOJ:
create override marker for remand case on probation

# 🚨 READ THIS 🚨

Does this change...
* [ ] 🚩 Need Feature flagged

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAPIController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAPIController.kt
@@ -7,6 +7,7 @@ import io.swagger.v3.oas.annotations.responses.ApiResponse
 import io.swagger.v3.oas.annotations.responses.ApiResponses
 import io.swagger.v3.oas.annotations.security.SecurityRequirement
 import io.swagger.v3.oas.annotations.tags.Tag
+import org.slf4j.LoggerFactory
 import org.springframework.data.repository.findByIdOrNull
 import org.springframework.http.HttpStatus.MOVED_PERMANENTLY
 import org.springframework.http.ResponseEntity
@@ -108,8 +109,13 @@ class ProbationAPIController(
       return
     }
 
-    require(defendant.overrideMarker == null && offender.overrideMarker == null) {
-      "Cannot assign include override marker to records with existing override markers"
+    if (defendant.overrideMarker != null || offender.overrideMarker != null) {
+      log.warn(
+        "Skipping include override markers for defendantId {} and offenderCrn {} due to existing override markers.",
+        defendant.defendantId,
+        offender.crn,
+      )
+      return
     }
 
     overrideService.systemInclude(defendant, offender)
@@ -125,4 +131,8 @@ class ProbationAPIController(
   private fun <T : Any> respondWithRedirect(crn: String): ResponseEntity<T> = ResponseEntity.status(MOVED_PERMANENTLY)
     .location(URI("/person/probation/$crn"))
     .build()
+
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAPIController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAPIController.kt
@@ -7,7 +7,6 @@ import io.swagger.v3.oas.annotations.responses.ApiResponse
 import io.swagger.v3.oas.annotations.responses.ApiResponses
 import io.swagger.v3.oas.annotations.security.SecurityRequirement
 import io.swagger.v3.oas.annotations.tags.Tag
-import org.slf4j.LoggerFactory
 import org.springframework.data.repository.findByIdOrNull
 import org.springframework.http.HttpStatus.MOVED_PERMANENTLY
 import org.springframework.http.ResponseEntity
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAPIController.kt
```diff
@@ -91,21 +94,45 @@ class ProbationAPIController(
...
   }
 
   private fun retrieveDefendant(defendantId: String): PersonEntity = personRepository.findByDefendantId(defendantId) ?: throw ResourceNotFoundException(defendantId)
 
+  private fun assignOverrideMarker(defendant: PersonEntity, offender: PersonEntity) {
+    if (recordsAreIncluded(defendant, offender)) {
+      return
+    }
+
+    if (defendant.overrideMarker != null || offender.overrideMarker != null) {
```
JunaedAhmedMOJ: Not sure if this specific scenario can happen but what if a CP record has an override scope with another remand record already?

JunaedAhmedMOJ:
Need to discuss if we want to recluster

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationAPIController.kt
JunaedAhmedMOJ: create a seperate handler for private methods, 
and do recluster after assigning the override marker as it is a field that triggers recluster on process person

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/probation/ProbationOverrideHandler.kt
```diff
@@ -0,0 +1,42 @@
...
+  private val overrideService: OverrideService,
+  private val personRepository: PersonRepository,
+  private val reclusterService: ReclusterService,
+) {
+  fun assignIncludeOverrideAndRecluster(defendant: PersonEntity, offender: PersonEntity) {
+    if (recordsAreIncluded(defendant, offender)) {
+      return
+    }
+
+    if (defendant.overrideMarker != null || offender.overrideMarker != null) {
```
markreesmoj: Override markers are not simply a 1:1 relationship.
I can't find any documentation on this but as I understand it, records in the person table have 0 or 1 override marker

Records can be included or excluded from more than one cluster.

This is achieved by use of override scope - [look at the code in here](https://github.com/ministryofjustice/hmpps-person-record/blob/180d458b8828c8432de1745df618a2428f49559b/src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/jpa/entity/PersonEntity.kt#L177)

So if either the offender or the defendant already has an override marker, that is no reason to stop processing. It is unlikely to happen, but if they do, we should still be able to include them

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/probation/ProbationOverrideHandler.kt
```diff
@@ -0,0 +1,42 @@
...
+        "Skipping include override markers for defendantId {} and offenderCrn {} due to existing override markers.",
+        defendant.defendantId,
+        offender.crn,
+      )
+      return
+    }
+
+    overrideService.systemInclude(defendant, offender)
+    personRepository.saveAll(listOf(defendant, offender))
+    offender.personKey?.let { reclusterService.recluster(offender) }
```
markreesmoj, line 22: we could recluster the defendant instead to avoid this conditional

src/main/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/handler/probation/ProbationOverrideHandler.kt
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  private fun recordsAreIncluded(defendant: PersonEntity, offender: PersonEntity): Boolean = defendant.overrideMarker != null &&
+    defendant.overrideMarker == offender.overrideMarker &&
```
markreesmoj, line 26: see above for a brief explanation of override marker logic

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationApiE2ETest.kt
```diff
@@ -591,7 +587,63 @@ class ProbationApiE2ETest : E2ETestBase() {
       }
 
       @Test
-      fun `should retain master defendant id on update on probation record`() {
+      fun `should assign include override marker to defendant and offender for create`() {
```
markreesmoj: I don't think we need this test

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationApiE2ETest.kt
```diff
@@ -591,7 +587,63 @@ class ProbationApiE2ETest : E2ETestBase() {
...
+
+        webTestClient.put()
+          .uri(probationApiUrl(defendantId))
+          .authorised(listOf(PROBATION_API_READ_WRITE))
+          .bodyValue(probationCase)
+          .exchange()
+          .expectStatus()
+          .isOk
+
+        val offender = awaitNotNull { personRepository.findByCrn(probationCase.identifiers.crn!!) }
```
markreesmoj, line 632: all we really need to assert here is that the defendant and offender are on the same personKey

src/test/kotlin/uk/gov/justice/digital/hmpps/personrecord/api/controller/probation/ProbationApiE2ETest.kt
```diff
@@ -621,11 +673,13 @@ class ProbationApiE2ETest : E2ETestBase() {
...
 
-        assertThat(offender.masterDefendantId).isEqualTo(defendant.masterDefendantId)
+        val updatedOffender = awaitNotNull { personRepository.findByCrn(probationCase.identifiers.crn!!) }
+        updatedOffender.assertIncluded(defendantRecord)
+        assertThat(updatedOffender.masterDefendantId).isNull()
       }
 
       @Test
-      fun `should set probation and court records that are on different clusters onto same cluster`() {
+      fun `should keep include override marker when probation and court records are on different clusters`() {
```
markreesmoj: This test is unrealistic. the defendant and the offender data would be identical. It can be deleted - the one above with minimal data which is not enough to match the records is proof that the linking  works

markreesmoj (dismissed):

lbevan (approved):
👍

